PR #1, opened 2022-12-15
Add Elastic APM (intake v2/v3) & OTLP inputs

axw:
Migrate code from apm-server/processor/stream. The code is _mostly_ unchanged - there are just a few significante changes:

- No more integration/approval tests in input/elasticapm. We no longer feed in canned data and check that the output events match golden files on disk. This can be done in the apm-server repo, in system tests. We're still doing it in input/otlp for now, but there's a TODO to change this.
- No benchmark tests in the Elastic APM input. These were doing a bit too much, similar to the integration tests mentioned above. We should replace them with microbenchmarks later.
- logp is replaced with zap.
- elastic-agent-libs/monitoring is removed from input/elasticapm; this will be handled in apm-server.
- The APIs have changed slightly.


axw:
@elastic/apm-server obviously it's not reasonable for you to review all 37000 lines of diff. Could you please have a look the package structure, and the API, and let me know if these look good?

axw:
Given the size of this change, and the ask to focus on the API, I think it's more practical to handle the review on the subsequent apm-server PR. I'll merge this and we can come back and make changes as needed.

PR #5, opened 2022-12-19
feat: translate otel log events to error events

kruskall:
Translate otel logs events that represent an error to error events. Based on the otel spec for recording exceptions, at least one of the following sets of attributes is required:

exception.type
exception.message

Additionally we want to map an app crash to an error event by checking the event.name.

Add test to validate the expected APMEvent output.

Closes https://github.com/elastic/apm-server/issues/9659


input/otlp/logs.go
```diff
@@ -109,12 +117,63 @@ func (c *Consumer) convertLogRecord(
 		}
 		event.Span.ID = spanID.HexString()
 	}
-	if attrs := record.Attributes(); attrs.Len() > 0 {
+	attrs := record.Attributes()
+	if isError(record) {
+		event.Processor = model.ErrorProcessor
+		setErrorLabels(attrs, &event)
+		updateErrorType(attrs, &event)
```
axw, line 124: Should we set Processor & Event.* in setErrorLabels as well?

input/otlp/logs.go
```diff
@@ -109,12 +117,63 @@ func (c *Consumer) convertLogRecord(
...
+	} else {
 		setLabels(attrs, &event)
 	}
 	return event
 }
 
+func updateErrorType(m pcommon.Map, event *model.APMEvent) {
```
axw, line 131: Should we remove the pcommon.Map param, since it's unused?

input/otlp/logs.go
```diff
@@ -109,12 +117,63 @@ func (c *Consumer) convertLogRecord(
...
+		}
+		return true
+	})
+
+	if exceptionMessage != "" || exceptionType != "" {
+		// Per OpenTelemetry semantic conventions:
+		//   `At least one of the following sets of attributes is required:
+		//   - exception.type
+		//   - exception.message`
+		event.Error = convertOpenTelemetryExceptionSpanEvent(
```
axw, line 164: Should we rename convertOpenTelemetryExceptionSpanEvent (convertOpenTelemetryException?), if it's not specific to span events any more?

input/otlp/logs.go
```diff
@@ -84,6 +85,13 @@ func (c *Consumer) convertInstrumentationLibraryLogs(
 	}
 }
 
+func isError(record plog.LogRecord) bool {
```
axw, line 88:
```suggestion
func isExceptionLogRecord(record plog.LogRecord) bool {
```

Let's make sure the function name makes sense at the package level, not just within the file. Also, I think it's important that we distinguish between "error logs" and "exception logs". "Error" might just imply that it's a log message with error log level, which we don't automatically turn into error events.

input/otlp/logs.go
axw: Perhaps we can get rid of this `setErrorLabels` function, and refactor `convertOpenTelemetryExceptionSpanEvent` so it iterates of a `pcommon.Map` for either logs or span events?

input/otlp/logs.go
```diff
@@ -109,12 +117,63 @@ func (c *Consumer) convertLogRecord(
...
+		updateErrorType(attrs, &event)
+	} else {
 		setLabels(attrs, &event)
 	}
 	return event
 }
 
+func updateErrorType(m pcommon.Map, event *model.APMEvent) {
+	event.Event.Kind = "event"
+	event.Event.Category = "device"
```
simitt, line 133: Should the category always be set, even if it is a _regular_ log exception? My understanding is that we always want to translate to error events if `semconv.AttributeExceptionType` or `semconv.AttributeExceptionMessage` is set, but only label as app crash (`event.name=crash` & `event.category=device`) when the custom `event.name=crash` is set.

model/event.go
```diff
@@ -43,13 +43,22 @@ type Event struct {
 	// source publishes more than one type of log or events (e.g. access log,
 	// error log), the dataset is used to specify which one the event comes from.
 	Dataset string
+
+	Kind string
+
+	Category string
+
+	Type string
```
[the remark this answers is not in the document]
kruskall: Thanks for spotting this! Updated :+1:

input/otlp/logs.go
kruskall: Good point! I removed `setErrorLabels` and reused the code in `setLabels`

input/otlp/logs.go
kruskall: this function has been removed!

input/otlp/logs.go
```diff
@@ -109,15 +110,54 @@ func (c *Consumer) convertLogRecord(
...
+		case semconv.AttributeExceptionMessage:
+			exceptionMessage = v.Str()
+		case semconv.AttributeExceptionStacktrace:
+			exceptionStacktrace = v.Str()
+		case semconv.AttributeExceptionType:
+			exceptionType = v.Str()
+		case semconv.AttributeExceptionEscaped:
+			exceptionEscaped = v.Bool()
+		default:
+			setLabel(replaceDots(k), event, ifaceAttributeValue(v))
```
kruskall, line 137: There seems to be a few inconsistencies with the usage of `replaceDots` and `setLabel` in the code. What's the correct behaviour ?

input/otlp/logs.go
```diff
@@ -109,15 +110,54 @@ func (c *Consumer) convertLogRecord(
...
+		event.Event.Kind = "event"
+		event.Event.Type = "error"
 	}
 	return event
 }
 
 func setLabels(m pcommon.Map, event *model.APMEvent) {
+	var exceptionEscaped bool
+	var exceptionMessage, exceptionStacktrace, exceptionType string
 	m.Range(func(k string, v pcommon.Value) bool {
```
kruskall, line 126: Should we include the `event.name` in the `APMEvent` labels ?

input/otlp/logs.go
```diff
@@ -109,15 +110,54 @@ func (c *Consumer) convertLogRecord(
...
+	setLabels(attrs, &event)
+	if event.Error != nil {
+		event.Processor = model.ErrorProcessor
+		event.Event.Kind = "event"
+		event.Event.Type = "error"
 	}
 	return event
 }
 
 func setLabels(m pcommon.Map, event *model.APMEvent) {
```
axw, line 123: I don't think we should change `setLabels`. Its job is meant to be set to set `labels.*` key/value pairs, and nothing else. Also, changing it means that the `setLabels(body.Map(), &event)` call on line 100/101 above would change behaviour.

Instead of doing this, how about we iterate over the attributes in this method, and just call `setLabel` for the attributes that aren't interpreted? That's what we do for spans.

input/otlp/logs.go
axw: We should be consistently replacing dots.

axw (changes requested):
Can you please add some unit tests for the new attribute handling?

input/otlp/logs.go
kruskall: Agree with the first part.

> Instead of doing this, how about we iterate over the attributes in this method, and just call setLabel for the attributes that aren't interpreted? That's what we do for spans.

Does "this method" refer to `setLabels` ? 

___

I've pushed some changes

axw: Sorry for being unclear, I meant in `convertLogRecord`.

kruskall: No worries, I've pushed some changes :+1:

input/otlp/logs.go
```diff
@@ -109,15 +110,64 @@ func (c *Consumer) convertLogRecord(
...
+		case semconv.AttributeExceptionMessage:
+			exceptionMessage = v.Str()
+		case semconv.AttributeExceptionStacktrace:
+			exceptionStacktrace = v.Str()
+		case semconv.AttributeExceptionType:
+			exceptionType = v.Str()
+		case semconv.AttributeExceptionEscaped:
+			exceptionEscaped = v.Bool()
+		case "event.name":
+			eventName = v.Str()
```
axw, line 131:
```suggestion
		case "event.name":
			eventName = v.Str()
		case "event.domain":
			eventDomain = v.Str()
```

input/otlp/logs.go
```diff
@@ -109,15 +110,64 @@ func (c *Consumer) convertLogRecord(
...
 	}
-	if attrs := record.Attributes(); attrs.Len() > 0 {
-		setLabels(attrs, &event)
+	attrs := record.Attributes()
+
+	var exceptionMessage string
+	var exceptionStacktrace string
+	var exceptionType string
+	var exceptionEscaped bool
+	var eventName string
```
axw, line 119:
```suggestion
	var eventName string
	var eventDomain string
```

axw (changes requested):

input/otlp/logs.go
```diff
@@ -109,15 +110,64 @@ func (c *Consumer) convertLogRecord(
...
+
+	if eventName != "" {
+		if event.Error == nil {
+			event.Error = &model.Error{}
+		}
+		if eventName == "crash" {
+			event.Event.Category = "device"
+		}
+		event.Error.Type = eventName
 	}
```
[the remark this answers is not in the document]
kruskall: Oh, I missed that with the refactor, thanks!

Good catch for the event.domain! :+1:

axw (approved):
LGTM!

PR #8, opened 2023-01-09
input/elasticapm: reset json-iterator correctly

axw:
When resetting a NDJSONStreamDecoder, make sure we don't read buffered data from a reused json-iterator decoder. This could happen if there is some junk data after an event object, but before the newline delimiter.

We fix this by not using json-iterator's Decoder type, but instead borrowing and releasing an Iterator with the buffered line.

Without the fix, we may see errors such as this:

```
=== RUN   TestIntake/Spans
    intake.go:46: 
                Error Trace:    /home/andrew/projects/apm-server/systemtest/intake.go:76
                                                        /home/andrew/projects/apm-server/systemtest/intake.go:46
                                                        /home/andrew/projects/apm-server/systemtest/intake_test.go:51
                Error:          Not equal: 
                                expected: 202
                                actual  : 400
                Test:           TestIntake/Spans
                Messages:       {"accepted":0,"errors":[{"message":"decode error: data read error: readObjectStart: expect { or n,","document":"{\"metadata\": {\"user\": {\"domain\": \"ldap://abc\", \"id\": \"123\", \"email\": \"s@test.com\", \"username\": \"john\"}, \"process\": {\"ppid\": 6789, \"pid\": 1234,\"argv\": [\"node\", \"server.js\"], \"title\": \"node\"}, \"system\": {\"platform\": \"darwin\", \"hostname\": \"prod1.example.com\", \"architecture\": \"x64\", \"container\": {\"id\": \"container-id\"}, \"kubernetes\": {\"namespace\": \"namespace1\", \"pod\": {\"uid\": \"pod-uid\", \"name\": \"pod-name\"}, \"node\": {\"name\": \"node-name\"}}}, \"labels\": {\"tag1\": \"label1\"}, \"service\": {\"name\": \"backendspans\", \"language\": {\"version\": \"8\", \"name\": \"ecmascript\"}, \"agent\": {\"version\": \"3.14.0\", \"name\": \"elastic-node\"}, \"environment\": \"staging\", \"framework\": {\"version\": \"1.2.3\", \"name\": \"Express\"}, \"version\": \"5.1.3\", \"runtime\": {\"version\": \"8.0.0\", \"name\": \"node\"}},\"cloud\":{\"account\":{\"id\":\"account_id\",\"name\":\"account_name\"},\"availability_zone\":\"cloud_availability_zone\",\"instance\":{\"id\":\"instance_id\",\"name\":\"instance_name\"},\"machine\":{\"type\":\"machine_type\"},\"project\":{\"id\":\"project_id\",\"name\":\"project_name\"},\"provider\":\"cloud_provider\",\"region\":\"cloud_region\",\"service\":{\"name\":\"lambda\"}}}}"}]}
```

This is due to a trailing "`" character in [apm-server/testdata/intake-v2/metricsets.json](https://github.com/elastic/apm-server/blob/7ac6d4dd062627a39c0dd079393a060ed2822766/testdata/intake-v2/metricsets.ndjson). I found this bug while moving integration tests out of internal/beater and into systemtest.

### DIFF
--- a/input/elasticapm/internal/decoder/stream_decoder_test.go
+++ b/input/elasticapm/internal/decoder/stream_decoder_test.go
@@ -112,3 +112,33 @@ func TestNDStreamReaderReadAhead(t *testing.T) {
 		}
 	}
 }
+
+func TestNDStreamDecoderTrailing(t *testing.T) {
+	dec := NewNDJSONStreamDecoder(strings.NewReader(`{"key":"value1"}junk`+"\n"), 100)
+
+	var out struct{ Key string }
+	require.NoError(t, dec.Decode(&out))
+	assert.Equal(t, "value1", out.Key)
+
+	dec.Reset(strings.NewReader(`{"key":"value2"}junk` + "\n"))
+	err := dec.Decode(&out)
+	require.NoError(t, err)
+	assert.Equal(t, "value2", out.Key)
+}
+
+func BenchmarkNDStreamDecoder(b *testing.B) {
+	b.RunParallel(func(pb *testing.PB) {
+		r := strings.NewReader("")
+		dec := NewNDJSONStreamDecoder(r, 100)
+		for pb.Next() {
+			dec.Reset(strings.NewReader("\"string\"\n"))
+			var out interface{}
+			if err := dec.Decode(&out); err != nil {
+				b.Fatal(err)
+			}
+			if out != "string" {
+				b.Fatalf("expected \"string\", got %v", out)
+			}
+		}
+	})
+}
--- a/input/elasticapm/internal/decoder/stream_decoder.go
+++ b/input/elasticapm/internal/decoder/stream_decoder.go
@@ -21,8 +21,6 @@ import (
 	"bufio"
 	"bytes"
 	"io"
-
-	jsoniter "github.com/json-iterator/go"
 )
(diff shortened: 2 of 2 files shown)


carsonip (approved):

input/elasticapm/internal/decoder/stream_decoder_test.go
```diff
@@ -112,3 +112,33 @@ func TestNDStreamReaderReadAhead(t *testing.T) {
 		}
 	}
 }
+
+func TestNDStreamDecoderTrailing(t *testing.T) {
+	dec := NewNDJSONStreamDecoder(strings.NewReader(`{"key":"value1"}junk`+"\n"), 100)
+
+	var out struct{ Key string }
+	require.NoError(t, dec.Decode(&out))
```
lahsivjar, line 120: Shouldn't this be an error since the line is not a valid JSON? Correct me if I am wrong, but my understanding was that in NDJSON we will delimit by `\n` and each line should be a valid JSON.

axw: Yes that would be better. We just decode the first JSON object off the line, and ignore anything after it. This is pre-existing behaviour, so I'll defer improving on that.

PR #16, opened 2023-02-10
Added representative count for OpenTelemetry spans

AlexanderWert:
Fixes elastic/apm#472

Based on [OTel's specification for probability sampling](https://opentelemetry.io/docs/reference/specification/trace/tracestate-probability-sampling/#p-value) we can use the p-value from the tracestate to derive the `adjusted count` (i.e. representative count).

This PR adds the representative count for OTel spans and transactions based on the p-value and, thus, solves elastic/apm#472 and the limitation documented on https://www.elastic.co/guide/en/apm/guide/current/open-telemetry-known-limitations.html#open-telemetry-metrics-limitations

### DIFF
--- a/input/otlp/traces.go
+++ b/input/otlp/traces.go
@@ -1077,9 +1077,6 @@ func getRepresentativeCountFromTracestateHeader(tracestace string) float64 {
 		}
 	}
 
-	if p < 0 {
-		return 1.0
-	}
 
 	if p > 62 {
 		return 0.0
--- a/input/otlp/traces.go
+++ b/input/otlp/traces.go
@@ -1064,7 +1064,11 @@ func schemeDefaultPort(scheme string) int {
 //	https://opentelemetry.io/docs/reference/specification/trace/tracestate-probability-sampling/#p-value
 //
 // to calculate the adjusted count (i.e. representative count)
+//
+// If the p-value is missing or invalid in the tracestate we assume
+// a sampling rate of 100% and a representative count of 1.
 func getRepresentativeCountFromTracestateHeader(tracestace string) float64 {
+	// Default p-value is 0, leading to a default representative count of 1.
 	var p uint64 = 0
 
 	otValue := getValueForKeyInString(tracestace, "ot", ',', '=')
@@ -1076,7 +1080,6 @@ func getRepresentativeCountFromTracestateHeader(tracestace string) float64 {
 			p, _ = strconv.ParseUint(pValue, 10, 6)
 		}
 	}
-
 
 	if p > 62 {
 		return 0.0
--- a/input/otlp/traces.go
+++ b/input/otlp/traces.go
@@ -1065,7 +1065,7 @@ func schemeDefaultPort(scheme string) int {
 //
 // to calculate the adjusted count (i.e. representative count)
 func getRepresentativeCountFromTracestateHeader(tracestace string) float64 {
-	var p float64 = 0.0
+	var p uint64 = 0
 
 	otValue := getValueForKeyInString(tracestace, "ot", ',', '=')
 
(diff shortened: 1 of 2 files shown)


input/otlp/traces.go
```diff
@@ -1056,3 +1057,55 @@ func schemeDefaultPort(scheme string) int {
...
+func getRepresentativeCountFromTracestateHeader(tracestace string) float64 {
+	var p float64 = 0.0
+
+	otValue := getValueForKeyInString(tracestace, "ot", ',', '=')
+
+	if otValue != "" {
+		pValue := getValueForKeyInString(otValue, "p", ';', ':')
+
+		if pValue != "" {
+			p, _ = strconv.ParseFloat(pValue, 64)
```
axw, line 1076: The spec states:

> Both fields are propagated via the OpenTelemetry tracestate under the ot vendor tag using the rules for [tracestate handling](https://opentelemetry.io/docs/reference/specification/trace/tracestate-handling/). Both fields are represented as unsigned decimal integers requiring at most 6 bits of information.

Should we use `strconv.ParseUint(pValue, 10, 6)` then? This would ignore spurious tracestates, as the spec says we should.

input/otlp/traces_test.go
```diff
@@ -99,21 +99,48 @@ func TestOutcome(t *testing.T) {
...
 }
 
+// func TestRepresentativeCountFromTracestateHeader(t *testing.T) {
+// 	assert.Equal(t, 1.0, otlp.GetRepresentativeCountFromTracestateHeader("invalid_trace_state"))
+// 	assert.Equal(t, 4.0, otlp.GetRepresentativeCountFromTracestateHeader("es=s:0.5,ot=p:2;r:62;k1:13,xy=w"))
+// 	assert.Equal(t, 256.0, otlp.GetRepresentativeCountFromTracestateHeader("esx:0.5,ot=p:8;r:62;k1:13,xy=w"))
+// 	assert.Equal(t, math.Pow(2.0, 62), otlp.GetRepresentativeCountFromTracestateHeader("ot=p:62"))
+// 	assert.Equal(t, 0.0, otlp.GetRepresentativeCountFromTracestateHeader("ot=p:63"))
+// }
+
```
axw, line 109:
```suggestion
```

axw (commented):
Looks good! Just a couple of little things.

input/otlp/traces.go
```diff
@@ -1056,3 +1057,55 @@ func schemeDefaultPort(scheme string) int {
...
+		pValue := getValueForKeyInString(otValue, "p", ';', ':')
+
+		if pValue != "" {
+			p, _ = strconv.ParseUint(pValue, 10, 6)
+		}
+	}
+
+	if p < 0 {
+		return 1.0
+	}
```
axw, line 1082:
```suggestion
```

input/otlp/traces.go
```diff
@@ -1056,3 +1057,55 @@ func schemeDefaultPort(scheme string) int {
...
+func getRepresentativeCountFromTracestateHeader(tracestace string) float64 {
+	var p uint64 = 0
+
+	otValue := getValueForKeyInString(tracestace, "ot", ',', '=')
+
+	if otValue != "" {
+		pValue := getValueForKeyInString(otValue, "p", ';', ':')
+
+		if pValue != "" {
+			p, _ = strconv.ParseUint(pValue, 10, 6)
```
axw, line 1076: The lint issue below raises a new question for me: if the "p" key is missing or invalid from tracestate, then `p` here will be zero. That leads to a representative count of 1 (2**0). Is that intentional? I guess that's the right default, since otherwise there should not have been a span event sent. Worth adding a comment?

axw: > it's not making it worse to the previous situation, right?

Agreed, thanks for the update

axw (approved):

felixbarny:
Should we use that p-value in our agents as well in favor of propagating via the vendor-specific [`es` tracestate vendor key](https://github.com/elastic/apm/blob/main/specs/agents/tracing-sampling.md#propagation)?

axw:
@felixbarny the OTel spec is experimental, so IMO we shouldn't do that just yet. But probably eventually, yes. It would be a little more complicated than what we do, as the p-value represents power of 2 fractions. So the agents would need to be updated to set p-values probabilistically, per the spec:

> The ConsistentProbabilityBased sampler is required to support probabilities that are not exact powers of two. To do so, implementations are required to select between the nearest powers of two probabilistically. For example, 5% sampling can be achieved by selecting 1/16 sampling 60% of the time and 1/32 sampling 40% of the time.

PR #18, opened 2023-02-13
[poc] add agent telemetry

SylvainJuge:


input/elasticapm/internal/modeldecoder/generator/cmd/main.go
```diff
@@ -61,8 +61,8 @@ func generateV2() {
 	if err != nil {
 		panic(err)
 	}
-	generateCode(p, parsed, []string{"metadataRoot", "errorRoot", "metricsetRoot", "spanRoot", "transactionRoot", "logRoot"})
-	generateJSONSchema(p, pkg, parsed, []string{"metadata", "errorEvent", "metricset", "span", "transaction"})
+	generateCode(p, parsed, []string{"metadataRoot", "errorRoot", "metricsetRoot", "spanRoot", "transactionRoot", "logRoot", "telemetryRoot"})
+	generateJSONSchema(p, pkg, parsed, []string{"metadata", "errorEvent", "metricset", "span", "transaction", "telemetry"})
```
SylvainJuge, line 65: @carsonip if you can add this to https://github.com/elastic/apm-data/pull/17 that would be awesome as it's a bit tricky to guess where and how to add root objects.

input/elasticapm/internal/modeldecoder/v2/decoder.go
```diff
@@ -1329,6 +1366,11 @@ func mapToTransactionModel(from *transaction, event *model.APMEvent) {
 	}
 }
 
+func mapToTelemetryModel(from *telemetry, event *model.APMEvent) {
+	event.Processor = model.TelemetryProcessor
+	// TODO : finish mapping
+}
```
SylvainJuge, line 1372: As we might not map this data to an actual persisted model, this might be a good place to send the data directly to the telemetry endpoint if we can do that directly from apm-data.

SylvainJuge:
Closing as we are taking another approach directly within apm-server, see https://github.com/elastic/apm-server/pull/10272

input/elasticapm/internal/modeldecoder/generator/cmd/main.go
carsonip: ah sorry I missed this message. Adding root objects is slightly harder since it requires changing `processor.go`. Let's discuss in #17 whether it should go into the PR

PR #23, opened 2023-03-06
BREAKING CHANGE: add serviceName to app logs dataset

kyungeunni:
BREAKING CHANGE
part of https://github.com/elastic/apm-server/issues/10291

Add service name to app logs dataset to create app logs datastream per service.

### DIFF
--- a/model/modelprocessor/datastream.go
+++ b/model/modelprocessor/datastream.go
@@ -26,9 +26,10 @@ import (
 )
 
 const (
-	logsType       = "logs"
-	appLogsDataset = "apm.app"
-	errorsDataset  = "apm.error"
+	logsType              = "logs"
+	logDefaultServiceName = "unknown"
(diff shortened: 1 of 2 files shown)


model/modelprocessor/datastream.go
```diff
@@ -87,6 +88,19 @@ func isRUMAgentName(agentName string) bool {
 	return false
 }
 
+func getAppLogsDataset(event *model.APMEvent) string {
+	serviceName := event.Service.Name
+	if serviceName == "" {
+		serviceName = logDefaultServiceName
+	}
```
kyungeunni, line 95: Even though ServiceName will be always supplied by the agents, added a safeguard here as I observed the `model.APMEvent` can omit the ServiceName and it can create datastreams like `logs-apm.app.-*`. But if we are confident that this will never happen, we can skip this.

PR #28, opened 2023-04-19
added processing for mobile lifecycle events

bryce-b:


axw:
> New events using the otel event API aren't showing up in discover. Could this be the problem?

I have no idea. Can you log the OTLP payload that leads to data loss?

> I worry that any custom event that is generated through the event api isn't being stored. Is this concern valid?

Span events should be stored as logs. We do have some tests that show this working with the OTel Go SDK.

bryce-b:
I added some changes to this PR to split otel events into a different data stream. Events are more critical than logs, and they may be drowned out in the logs data stream.

kruskall:
Should we add some tests ?

bryce-b:
We decided not to move forward with creating a new data stream for log events. The paired PR apm_server/#10730 has been closed. I've refactored this PR to simply handle Open-Telemetry log events in a generic way.

axw:
@bryce-b I don't really understand the change in this PR. Can you elaborate on the rationale?

[`event.type`](https://www.elastic.co/guide/en/ecs/current/ecs-allowed-values-event-type.html), like other `event.*` categorisation fields, has a limited set of allowed values. So I'm not sure that we should be passing through the `event.name` value specified by the client.

I also don't know that we should infer `event.kind` from `event.domain == device`. Couldn't there theoretically be other non-event logs coming from devices?

bryce-b:
@axw I think I'm just confused about the elastic spec vs the open-telemetry spec. I was trying to extrapolate a more generic handling of the [open-telemetry event API ](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/event-api.md) based on how the crash events are handled (which uses the same otel API). 
In otel, logs with `event.domain` & `event.name` are a special event type.  It looks like putting the name of the event into `event.type` isn't going to work, but we'll need a place to put `event.name` as it will be an identifier expected by customers. 

In otel, the `event.domain` field identifies whether the log is from the browser, mobile device, or kubernetes. Maybe it's more appropriate to ignore `event.domain`?

axw:
> In otel, logs with event.domain & event.name are a special event type.

I see, thanks for the explanation. I think we should make the "is an event" inference rules a bit more general, and base it on the presence of both `event.domain` and `event.name`.

> It looks like putting the name of the event into event.type isn't going to work, but we'll need a place to put event.name as it will be an identifier expected by customers.

Maybe [`event.action`](https://www.elastic.co/guide/en/ecs/master/ecs-event.html#field-event-action)? With the merger of ECS and OTel semantic conventions we should probably set `event.name` eventually, but the rules aren't yet defined. WDYT @AlexanderWert?

So my proposal would be:
 - if both `event.domain` and `event.name` are set, then we infer that it's an event log and set `event.kind: event`
 - for such event logs, if `event.domain: device` then we set `event.category: device`
 - for such event logs, we set `event.action: <event.name>`

AlexanderWert:
> Maybe [event.action](https://www.elastic.co/guide/en/ecs/master/ecs-event.html#field-event-action)? With the merger of ECS and OTel semantic conventions we should probably set event.name eventually, but the rules aren't yet defined. WDYT @AlexanderWert?

The handling `event.*` fields in the context of ECS/OTel merger has not been discussed, yet.
So, let's start with @axw's proposal. Most likely we will need to adapt later (when the corresponding SemConv are stabilized) anyways.

bryce-b:
@axw could you re-review this pr?

input/otlp/logs.go
```diff
@@ -135,6 +135,8 @@ func (c *Consumer) convertLogRecord(
 			eventDomain = v.Str()
 		case "session.id":
 			event.Session.ID = v.Str()
+		case "lifecycle.state":
+			setLabel(k, &event, ifaceAttributeValue(v))
```
axw, line 139: Why do we need special handling for `lifecycle.state`?

input/otlp/logs.go
```diff
@@ -152,14 +154,19 @@ func (c *Consumer) convertLogRecord(
...
 		if event.Error == nil {
 			event.Error = &model.Error{}
 		}
 		event.Error.Type = "crash"
-	}
-	if eventDomain == "device" {
-		event.Event.Category = "device"
+		event.Event.Kind = ""
+		event.Event.Action = ""
 	}
```
axw, line 170: Can you update this to follow https://github.com/elastic/apm-data/pull/28#issuecomment-1539268769?

> - if both `event.domain` and `event.name` are set, then we infer that it's an event log and set `event.kind: event`
> - for such event logs, if `event.domain: device` then we set `event.category: device`
> - for such event logs, we set `event.action: <event.name>`

axw (changes requested):

input/otlp/logs.go
bryce-b: Maybe I'm a little confused about how new attributes get added to ECS, but this was to prevent the `label` prefix from being prepended.

input/otlp/logs.go
axw: Sorry! My eyes were drawn to the changes in the crash code block and I didn't pay close attention to the other code block.

IIUC, "crash" events are only going to come from mobile devices, so I think we should combine them and avoid setting the Kind & Action fields only to clear them out afterwards.

```go
if eventDomain == "device" && eventName != "" {
	event.Event.Category = "device"
	if eventName == "crash" {
		if event.Error == nil {
			event.Error = &model.Error{}
		}
		event.Error.Type = "crash"
	} else {
		event.Event.Kind = "event"
		event.Event.Action = eventName
	}
}
```

WDYT?

input/otlp/logs.go
axw: Nope, you're calling `setLabel` so it will end up as `labels.lifecycle.state`.

At the moment there's several things that are needed to add a top-level field, and map data to it:
 1. Follow https://github.com/elastic/ecs/blob/main/CONTRIBUTING.md to propose the addition to ECS. I guess OTel semantic conventions now?
 2. Add the field to relevant data streams in https://github.com/elastic/apm-server/tree/main/apmpackage
 3. Update https://github.com/elastic/apm-data/tree/main/model and https://github.com/elastic/apm-data/tree/main/model/proto to add the new field to our model types.
 4. Update this code to map the OTel attribute to the field in our model.

Perhaps move this to a new PR once step 1 has happened?

PR #30, opened 2023-04-25
Bring default RepresentativeCount for Spans in line with Transactions

Mpdreamz:
Relates: https://github.com/elastic/apm-server/pull/9458

I think we did not do this for spans because we assume new agents always
send `sample_rate` if `sample_rate > 0`

However this may not be true for non root transactions where the trace
was started by an external tool.

> For non-root transactions the agent MUST parse incoming tracestate headers to identify the es entry and extract the s attribute.
> The s attribute value should be used to populate the sample_rate field of transactions and spans.
> If there is no tracestate or no valid es entry with an s attribute, then the agent MUST omit sample_rate from non-root transactions and their spans.

See: https://github.com/elastic/apm/blob/main/specs/agents/tracing-sampling.md#propagation

This in turn means we won't collect span metrics for these transactions

### DIFF
--- a/input/elasticapm/internal/modeldecoder/v2/decoder.go
+++ b/input/elasticapm/internal/modeldecoder/v2/decoder.go
@@ -1063,8 +1063,17 @@ func mapToSpanModel(from *span, event *model.APMEvent) {
 	if from.ParentID.IsSet() {
 		event.Parent.ID = from.ParentID.Val
 	}
-	if from.SampleRate.IsSet() && from.SampleRate.Val > 0 {
-		out.RepresentativeCount = 1 / from.SampleRate.Val
+	if from.SampleRate.IsSet() {
+		if from.SampleRate.Val > 0 {
+			out.RepresentativeCount = 1 / from.SampleRate.Val
+		}
+	} else {
+		// NOTE: we don't know the sample rate, so we need to make an assumption.
+		//
+		// Agents never send spans for non-sampled transactions, so assuming a
+		// representative count of 1 (i.e. sampling rate of 100%) may be invalid; but
+		// this is more useful than producing no metrics at all (i.e. assuming 0).
+		out.RepresentativeCount = 1
 	}
 	if len(from.Stacktrace) > 0 {
 		out.Stacktrace = make(model.Stacktrace, len(from.Stacktrace))
--- a/input/elasticapm/internal/modeldecoder/v2/span_test.go
+++ b/input/elasticapm/internal/modeldecoder/v2/span_test.go
@@ -195,11 +195,11 @@ func TestDecodeMapToSpanModel(t *testing.T) {
 		input.SampleRate.Set(0.25)
 		mapToSpanModel(&input, &out)
 		assert.Equal(t, 4.0, out.Span.RepresentativeCount)
-		// sample rate is not set
+		// sample rate is not set, default representative count should be 1
 		out.Span.RepresentativeCount = 0.0
 		input.SampleRate.Reset()
 		mapToSpanModel(&input, &out)
-		assert.Equal(t, 0.0, out.Span.RepresentativeCount)
+		assert.Equal(t, 1.0, out.Span.RepresentativeCount)
 		// sample rate is set to 0
 		input.SampleRate.Set(0)
 		mapToSpanModel(&input, &out)
--- a/input/elasticapm/internal/modeldecoder/v2/transaction_test.go
+++ b/input/elasticapm/internal/modeldecoder/v2/transaction_test.go
@@ -643,7 +643,8 @@ func TestDecodeMapToTransactionModel(t *testing.T) {
 		var out model.APMEvent
 		mapToSpanModel(&input, &out)
 		assert.Equal(t, model.Span{
-			Type: "unknown",
+			Type:                "unknown",
+			RepresentativeCount: 1,
 		}, *out.Span)
 	})
 }


axw (changes requested):
Can you please open an issue in elastic/apm to discuss this? The change in this PR is subtly breaking, and there are other options:
 - release a new major version of the RUM agent which always sends `tracestate` by default
 - set `trace_continuation_strategy` to `restart_external` by default

I'm not sure if those are reasonable, but we should discuss them.

felixbarny:
@axw are we assuming a representative count for transaction metrics if the `sample_rate` is missing? If so, why aren't we doing the same for span destination metrics? If not, I suppose we wouldn't be able to create any transaction metrics so the service might not even show up in the UI.

axw:
> @axw are we assuming a representative count for transaction metrics if the sample_rate is missing?

We set representative count to 1 when the sample_rate hasn't been sent: https://github.com/elastic/apm-data/blob/d1edc48bb5e38a58921a82c2c141c89d93eb0e6b/input/elasticapm/internal/modeldecoder/v2/decoder.go#L1271

> If so, why aren't we doing the same for span destination metrics?

Before we introduced tracestate & sample rate propagation, agents _always_ sent a transaction document. We're not assuming anything, we know that if sample_rate isn't set, then each transaction event is worth exactly 1. That is not the case for spans; spans are/were never sent for non-sampled transactions.

felixbarny:
> we know that if sample_rate isn't set, then each transaction event is worth exactly 1

I'm not sure if that assumption also holds true for OTel agents that don't send unsampled transactions and also don't send a sample_rate.

axw:
>> we know that if sample_rate isn't set, then each transaction event is worth exactly 1
>
> I'm not sure if that assumption also holds true for OTel agents that don't send unsampled transactions and also don't send a sample_rate.

It doesn't hold for OTel, I was only talking about intakev2.

OTel agents behave differently do Elastic APM agents, so OTLP is handled differently. For OTel we assume representative count of 1 for a missing p-value: https://github.com/elastic/apm-data/blob/main/input/otlp/traces.go#L1068-L1069. In lieu of a p-value (which will always be the case until the experimental sampling rate propagation spec is implemented), we _must_ make an assumption.

input/elasticapm/internal/modeldecoder/v2/decoder.go
```diff
@@ -1063,8 +1063,12 @@ func mapToSpanModel(from *span, event *model.APMEvent) {
...
 		event.Parent.ID = from.ParentID.Val
 	}
-	if from.SampleRate.IsSet() && from.SampleRate.Val > 0 {
-		out.RepresentativeCount = 1 / from.SampleRate.Val
+	if from.SampleRate.IsSet() {
+		if from.SampleRate.Val > 0 {
+			out.RepresentativeCount = 1 / from.SampleRate.Val
+		}
+	} else {
+		out.RepresentativeCount = 1
```
axw, line 1071:
```suggestion
		// NOTE: we don't know the sample rate, so we need to make an assumption.
		//
		// Agents never send spans for non-sampled transactions, so assuming a
		// representative count of 1 (i.e. sampling rate of 100%) may be invalid; but
		// this is more useful than producing no metrics at all (i.e. assuming 0).
		out.RepresentativeCount = 1
```

axw (approved):

felixbarny:
> I still think the other options I mentioned at https://github.com/elastic/apm-data/pull/30#pullrequestreview-1400911836 are worth considering, but I think we can go ahead with this.

I don't think these options are viable.

> release a new major version of the RUM agent which always sends tracestate by default

This doesn't fix the issue where the upstream service is monitored by OTel and doesn't propagate the sampling rate. But generally, I agree that we should do that anyway.

> set trace_continuation_strategy to restart_external by default

This would restart all traces from OTel agents so I don't think we'd want to do that by default as it would break interoperability with OTel agents.

axw:
> This would restart all traces from OTel agents so I don't think we'd want to do that by default as it would break interoperability with OTel agents.

Do they have to be in the same trace? What breaks if they're two traces that are linked?

felixbarny:
The biggest would be to be able to propagate the sampling decision as restart always implies to re-sample. But also things like service maps, log correlation, and being able to view the end-to-end trace in the waterfall, although these arguably have more to do with limitations on our support for span links.

gregkalapos:
> > set trace_continuation_strategy to restart_external by default
>
> This would restart all traces from OTel agents so I don't think we'd want to do that by default as it would break interoperability with OTel agents.

I'd add here that it's not only OTel - it'll break all traces where the upstream service is not instrumented by our agent but any other agent. So in sum, I feel if we'd do this, then our default would go against the W3C TraceParent spec (something like this is already mentioned in the [problem statement](https://www.w3.org/TR/trace-context/#problem-statement) section)

Mpdreamz:
squashed and signed the commit so this should be good to merge. 

> Can you please open an issue in elastic/apm to discuss this? The change in this PR is subtly breaking, and there are other options.

Will do for the next one @axw, apologies for leading with a solution rather than fully detailing the problem.

> This doesn't fix the issue where the upstream service is monitored by OTel and doesn't propagate the sampling rate. But generally, I agree that we should do that anyway.

This was indeed the main motivation of this PR, thanks @felixbarny for highlighting that.

axw:
:heart: thanks @Mpdreamz!

PR #33, opened 2023-05-09
model: Only write DSS source for metricsSets

marclop:
Removes the `modelprocessor.DroppedSpansStatsDiscarder` in favor of only writing the `dropped_span_stats` source when `MetricsetProcessor` is set as the `event.Processor`.


axw (approved):
I'm a little bit of two minds about this.

I'd like to make the JSON encoding as mechanical and dumb as possible, so we can generate it. Perhaps that shouldn't be based off the model types directly though, but we should instead have an intermediate step where we convert from the model types to some other types which have generated JSON encoding. That intermediate step would copy across the DSS only for metricsets.

Having the logic of dropping DSS external to the types is also not ideal, so I think this change is net positive.

Issue #36, opened 2023-05-12
Introduce protobuf definitions for model types


axw:
For [several reasons](https://github.com/elastic/apm-server/issues/4120) we would like to define an efficient, stable, binary encoding for model types. e.g. we would use this for storing events in Badger for tail-based sampling. These would be much faster to encode/decode, and more importantly will have strong stability guarantees

To achieve the above, we will  define our intermediate, in-memory/on-disk, model types in protobuf -- this will be the source of truth. We'll take a phased approach to this, given that the existing types are used all across apm-server, and making a big-bang change would carry a significant amount of risk of introducing bugs.

With https://github.com/elastic/apm-data/pull/35 merged, we have created a cleaner separation between the model types and the way they are encoded to JSON. The model types no longer have to directly map to the final document structure, though for our sanity we should probably keep them close. The model types do not need to be ECS-compliant, and we can instead evolve the JSON encoding over time without changing the model types.

**Phase 1**

 1. introduce protobuf definitions (probably worth basing off @marclop's work in https://github.com/elastic/apm-ingest-queueless)
 2. generate Go types from protobuf (into `model/modelpb` or something like that)
 3. introduce code for JSON encoding protobuf types by transforming to `internal/modeljson` types, like we're doing with `model.APMEvent` now
 4. introduce code for mapping model events to the protobuf-generated types; remove the code for translating from model types to modeljson, and instead translate model types to protoc-generated types, then protoc-generated types to modeljson

**Phase 2**

 1. Update apm-server to use protoc-generated Go types everywhere, and removing handwritten model types
 2. Update apm-data to decode events directly into protobuf-generated types, and remove the old model types

PR #39, opened 2023-05-25
feat: generate protobuf go models 

kruskall:
Add POSIX shell scripts to install protobuf and generate go models: protobuf is downloaded to the `/tmp` directory and installed in `tools/bin` along with `protoc-gen-go`. Go models are generated in the `model/modelpb` directory.

Make tasks are added for ease of use.


Makefile
```diff
@@ -2,7 +2,10 @@ test:
 	go test -v ./...
 
 fmt:
-	go run golang.org/x/tools/cmd/goimports@v0.3.0 -w .
+	go run golang.org/x/tools/cmd/goimports@v0.3.0 -w ./input
+	go run golang.org/x/tools/cmd/goimports@v0.3.0 -w ./model/modelprocessor
+	go run golang.org/x/tools/cmd/goimports@v0.3.0 -w ./model/internal
+	go run golang.org/x/tools/cmd/goimports@v0.3.0 -w ./model/*.go
```
axw, line 8: What's the reason for not formatting the generated code as well?

axw (commented):
Looks good. I think we should also use [vtprotobuf](https://github.com/planetscale/vtprotobuf), like at https://github.com/elastic/apm-ingest-service/blob/a0bcd8e2c2039471672383c5ba1c6f7ddc6bf5a4/Makefile#L42-L72

Makefile
kruskall: `goimports` is putting stdlib import first which is conflicting with the output generated from protobuf.

To give an example with `os.pb.go`, here's a diff of the file after generation and after `goimports` was applied:

```patch
diff --git a/model/modelpb/os.pb.go b/model/modelpb/os.pb.go
index 021e22f..25f0ce6 100644
--- a/model/modelpb/os.pb.go
+++ b/model/modelpb/os.pb.go
@@ -24,10 +24,11 @@
 package modelpb
 
 import (
-       protoreflect "google.golang.org/protobuf/reflect/protoreflect"
-       protoimpl "google.golang.org/protobuf/runtime/protoimpl"
        reflect "reflect"
        sync "sync"
+
+       protoreflect "google.golang.org/protobuf/reflect/protoreflect"
+       protoimpl "google.golang.org/protobuf/runtime/protoimpl"
 )
 
 const (
```

axw: Why is that a problem? It should be fine to reorder to goimports style.

tools/generate-modelpb.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+TOOLS_DIR=$(dirname "$(readlink -f -- "$0")")
+
+PATH="${PATH}:${TOOLS_DIR}/bin" protoc --proto_path=./model/proto/ --go_out=. --go_opt=module=github.com/elastic/apm-data --go-vtproto_out=. --go-vtproto_opt=features=marshal+unmarshal+size,module=github.com/elastic/apm-data ./model/proto/*.proto
```
axw, line 5:
```suggestion
PATH="${PATH}:${TOOLS_DIR}/bin" protoc --proto_path=./model/proto/ --go_out=. --go_opt=module=github.com/elastic/apm-data --go-vtproto_out=. --go-vtproto_opt=features=marshal+unmarshal+size+pool,module=github.com/elastic/apm-data ./model/proto/*.proto
```

axw (approved):
LGTM! I think we want to generate pooling support for vtproto as well.

PR #42, opened 2023-05-26
feat: introduce code for JSON encoding protobuf types

kruskall:
Introduce code for JSON encoding protobuf types

I tried to keep them close to the original apmevent `toModelJSON` but it should be fairly contained

Blocked by https://github.com/elastic/apm-data/pull/41


model/modelpb/client.pb.json.go
```diff
@@ -0,0 +1,36 @@
...
+func (c *Client) toModelJSON(out *modeljson.Client) {
+	*out = modeljson.Client{
+		Domain: c.Domain,
+		Port:   int(c.Port),
+	}
+	if c.Ip != "" {
+		if _, err := netip.ParseAddr(c.Ip); err == nil {
+			out.IP = c.Ip
+		}
+	}
```
axw, line 35: It doesn't seem ideal that we need to parse the IP here, to convert it into a netip.Addr, only to then convert it back into a string.

I wonder if we should introduce a more efficient IP type in our protobuf, along these lines: https://stackoverflow.com/questions/51395679/what-is-the-most-efficient-protobuf-type-in-c-for-storing-ipv4-or-ipv6-addre

WDYT? Is it overkill?

model/modelpb/span.pb.json.go
```diff
@@ -0,0 +1,105 @@
...
+		Link:         db.Link,
+		RowsAffected: db.RowsAffected,
+		User:         modeljson.DBUser{Name: db.UserName},
+	}
+}
+
+func (c *Composite) toModelJSON(out *modeljson.SpanComposite) {
+	sumDuration := time.Duration(c.Sum * float64(time.Millisecond))
+	*out = modeljson.SpanComposite{
+		CompressionStrategy: c.CompressionStrategy.String(),
```
axw, line 40: The value should be "exact_match" or "same_kind", I don't think this is right

axw (commented):
Looks good! I think the CompressionStrategy conversion is wrong, so that's a blocker, but the other things I mentioned are debatable.

While you're here, would you mind changing the `port` fields in protobuf from uint32 to uint16?

model/modelpb/client.pb.json.go
kruskall: No, I fully agree we should store it in a better way (bytes ?) but probably in a followup PR (I think there's a TODO about it)

This goal of `ParseAddr` here is to replicate the `ip.isValid()` call in the client model. Looking more at it, it seems that that's used to make sure that the ip is not a zero value so the if check might with the empty string might be enough here. WDYT ?

model/modelpb/apmevent.pb.json.go
```diff
@@ -71,33 +66,33 @@ func (e *APMEvent) toModelJSON(out *modeljson.Document) {
 		}
 	}
 
-	var transaction modeljson.Transaction
 	if e.Transaction != nil {
+		var transaction modeljson.Transaction
```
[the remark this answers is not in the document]
kruskall: Thank you for the feedback! 

I added benchmarks and there are quite a few allocations. I spent quite a while trying to fight the escape analysis but without much success. I'm trying to evalute how feasible it is to put together some code generation to automate the mapping.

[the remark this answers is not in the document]
kruskall: Thank you for the feedback!

I added benchmarks and there are quite a few allocations. I spent quite a while trying to fight the escape analysis but without much success. I'm trying to evalute how feasible it is to put together some code generation to automate the mapping but I'm afraid there's not much we can do to reduce allocation.

From what I could see from the output of escape analysis (`-m -m -m`) the variables are escaping to the heap because the struct field is a pointer (as opposed to `model.APMEvent` being composed of struct for the most part).
I tried playing around with the code but the benchmark is always reporting ~104 allocations.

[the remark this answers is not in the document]
kruskall: I tried another way that eliminates indirectly assigning the pointer and pre-assign them on the parent function so that they won't escape to the heap.
Allocations are down to ~59 and the mapping is definitely faster. Let me know what you think!

Note: tests are broken.

model/modelpb/client.pb.json.go
axw: It's fine as it is for now, but a followup would be good before we commit to keeping these protobufs stable.

axw:
>> While you're here, would you mind changing the port fields in protobuf from uint32 to uint16?
>
> Could you clarify this ? Protobuf doesn't have a uint16 AFAIK

Ah right, I had forgotten this. They will be varints, and the number of bytes will depend on the value. Nothing to do here then.

model/modelpb/apmevent.pb.json.go
```diff
@@ -0,0 +1,279 @@
...
+		Session:     &modeljson.Session{},
+		Process:     &modeljson.Process{},
+		Event:       &modeljson.Event{},
+	}
+	e.updateModelJSON(&out)
+	return ErrInvalidLength
+	//return out.MarshalFastJSON(w)
+}
+
+//go:noinline
```
axw, line 95: why would we not want to inline this?

model/modelpb/apmevent.pb.json.go
```diff
@@ -0,0 +1,279 @@
...
+func setNonZero[T comparable](to **T, from *T) {
+	if !isZero(*from) {
+		*to = from
+	}
+}
+
+func isZero[T comparable](t T) bool {
+	var zero T
+	return t == zero
+}
```
axw, line 279:
```suggestion
```

These don't seem to be used, can we remove them?

model/modelpb/apmevent.pb.json.go
```diff
@@ -0,0 +1,279 @@
...
+		},
+		URL:         &modeljson.URL{},
+		Log:         &modeljson.Log{},
+		Source:      &modeljson.Source{},
+		Client:      &modeljson.Client{},
+		Child:       &modeljson.Child{},
+		Destination: &modeljson.Destination{},
+		Session:     &modeljson.Session{},
+		Process:     &modeljson.Process{},
+		Event:       &modeljson.Event{},
```
axw, line 88: Won't this cause a lot of empty struct fields in the JSON output? It will also lead to unnecessary allocations.

model/modelpb/apmevent.pb.json.go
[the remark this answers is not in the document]
axw: > I tried another way that eliminates indirectly assigning the pointer and pre-assign them on the parent function so that they won't escape to the heap.
> Allocations are down to ~59 and the mapping is definitely faster. Let me know what you think!

Preassigning them doesn't seem ideal to me. Does the approach in `model.APMEvent.MarshalFastJSON` not work? The idea there is that we define the structs at the top level of the method, so they shouldn't escape.

model/modelpb/apmevent.pb.json.go
kruskall: Good catch! This is a leftover from my previous test with the mid-stack inliner. This is not used anyway as the function is too big to be inlined now. I've removed it! :+1:

model/stacktrace.go
```diff
@@ -57,11 +57,14 @@ func (s *StacktraceFrame) toModelJSON(out *modeljson.StacktraceFrame) {
...
 		Module:              s.Module,
 		Function:            s.Function,
-		Vars:                s.Vars,
 		LibraryFrame:        s.LibraryFrame,
 		ExcludeFromGrouping: s.ExcludeFromGrouping,
 	}
 
+	if len(s.Vars) != 0 {
+		out.Vars = s.Vars
+	}
```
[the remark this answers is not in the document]
kruskall: The test will fail without this. I guess it's mostly because of consistency: we are not copying over slice/maps if they are empty in other structs.

model/modelpb/apmevent.pb.json.go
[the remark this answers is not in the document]
kruskall: I've added a commit that changes back to the old approach (it can be reverted easily for testing).

Allocations went from 59 to 71 and cpu time was slightly slower.

I also added a benchmark for comparison with `model.APMEvent`.

model/modelpb/apmevent.pb.json.go
kruskall: > Won't this cause a lot of empty struct fields in the JSON output? 

This should be fixed, mapping will take care of setting them to nil.

> It will also lead to unnecessary allocations.

No, allocations decreased because these are all allocated on the stack. Before they were escaping to the heap so we had more allocations. See the benchmark above.

model/modelpb/apmevent.pb.json.go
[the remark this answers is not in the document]
kruskall: If the tradeoff is acceptable (increased allocations) we can keep it this way and try to improve later. I don't have a strong opinion on it. Let me know what you think

model/modelpb/apmevent.pb.json.go
axw: > This should be fixed, mapping will take care of setting them to nil.

:+1: I missed that we were setting them to nil.

> No, allocations decreased because these are all allocated on the stack. Before they were escaping to the heap so we had more allocations. See the benchmark above.

Ah, I think the issue is the nested objects. The outer types declared should be on the stack, but the inner ones won't be.

Let's go with preallocation for now then, so we can make progress.

model/modelpb/client.pb.json.go
kruskall: Opened https://github.com/elastic/apm-data/issues/47

model/modelpb/span.pb.json.go
kruskall: Fixed!

model/modelpb/apmevent.pb.json.go
kruskall: Good catch! Removed!

model/modelpb/apmevent.pb.json.go
kruskall: We discussed it and agree on a path forward for now!

model/apmevent.go
```diff
@@ -416,7 +416,7 @@ func (e *APMEvent) MarshalFastJSON(w *fastjson.Writer) error {
 		doc.Session = &session
 	}
 
-	return doc.MarshalFastJSON(w)
+	return net.ErrClosed
```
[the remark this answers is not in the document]
kruskall: This was added so that the benchmarks would only measure mapping performance, without running the marshaling logic.

It's a temporary solution to have more specialized benchmarks. We can remove it now.

kruskall:
Pushed a few more fixes I found while integrating this with APM Server (no PR yet). This should be ready for review again.

axw (approved):
Thank you!

PR #45, opened 2023-05-30
feat: introduce code for mapping model events to the protobuf-generated types

kruskall:
Add model->modelpb mapping logic.

Marshaling a model will now convert to modelpb under the hood. modelpb is then marshaled to json.

Update http headers to fix conversion error: structpb does not support `[]string`.


axw (approved):
nice :)

PR #53, opened 2023-06-09
Make span-type derivation from Otel data deterministic.

### LINKED ISSUE #43: Derivation of the `span.type` from OTel data is indeterministic
in OTel a single span can have a mix of attributes from different namespaces. For example, a span could have `db.*` attributes and at the same time `http.*` attributes. 

We use the logic of [this switch statement](https://github.com/elastic/apm-data/blob/0e8423572156bc44eea8b682c0dad1fa96d949cf/input/otlp/traces.go#L481) to determine the `foundSpanType` variable. This is done while iterating over all the attributes on a span. So basically the `last` attribute defines the actual `foundSpanType` and the corresponding mapping-logic below. This is indeterministic because we don't know the order of the span attributes.

We need a more explicit logic to derive the `span.type`.

JonasKunz:
Closes #43.

### DIFF
--- a/input/otlp/traces.go
+++ b/input/otlp/traces.go
@@ -228,7 +228,7 @@ func TranslateTransaction(
 		httpResponse   modelpb.HTTPResponse
 	)
 
-	var foundSpanType int
+	var isHTTP, isRPC, isMessaging bool
 	var message modelpb.Message
 
 	var samplerType, samplerParam pcommon.Value
@@ -255,7 +255,7 @@ func TranslateTransaction(
 		case pcommon.ValueTypeInt:
 			switch kDots {
 			case semconv.AttributeHTTPStatusCode:
-				foundSpanType = httpSpan
+				isHTTP = true
 				httpResponse.StatusCode = int32(v.Int())
 				http.Response = &httpResponse
 			case semconv.AttributeNetPeerPort:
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,3 +1,5 @@
 *.swp
+# IntelliJ module files (.idea folder, *.iml module files):
 .idea
+*.iml
```
marclop, line 4: These changes aren't related to this change, would you mind reverting them?

JonasKunz: Sure, I'll open a separate PR.

marclop (approved):
This LGTM. @axw do you mind a quick review given your OTEL expertise?

axw (commented):
Looks good, but IIANM we have another issue: if we set foundSpanType to databaseSpan (say), then none of the HTTP fields are set -- see line 369. I think that would matter for Elasticsearch, which provides an HTTP API for database operations.

I think we can address that while also simplifying a little bit by deferring the priority call until after the attribute processing. Something along these lines:

```go
var isHTTP, isDatabase, isRPC, isMessaging bool

// ... process attributes, setting zero or more of the above flags ...

if event.Transaction.Type  == "" {
    switch {
    case isMessaging:
        event.Transaction.Type = "messaging"
    case isRPC, isHTTP:
        event.Transaction.Type = "request"
    default:
        event.Transaction.Type = "unknown"
    }
}

if isHTTP {
    event.HTTP = http
    // set event.Event.Outcome, event.Transaction.Result, event.URL
}
```

WDYT?

axw (approved):
LGTM!

JonasKunz:
I set up commit signing and force-pushed a squashed commit without any additional changes. So should be good to merge now.

PR #54, opened 2023-06-12
Update gitignore for IntelliJ module files

JonasKunz:
Title says it all.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
+# vim
 *.swp
+# IntelliJ module files (.idea folder, *.iml module files):
 .idea
+*.iml
+# project build
 tools/build


dmathieu:
If other repositories are adding this, then 👍 
But IDE-specific things don't need to be added to a project's gitignore. They can be added to a global one: https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

basepi:
In my opinion, adding IDE-specific gitignore entries is not a problem, as long as they have comments separating them from language-specific gitignore entries. This is not a file where churn (or extra lines) really cost anything.

.gitignore
```diff
@@ -1,3 +1,5 @@
 *.swp
```
SylvainJuge, line 1:
```suggestion
# vim
*.swp
```

.gitignore
```diff
@@ -1,3 +1,5 @@
...
+# IntelliJ module files (.idea folder, *.iml module files):
 .idea
+*.iml
 tools/bin
```
SylvainJuge, line 5:
```suggestion
# project build
tools/bin
```

SylvainJuge (approved):

axw (approved):
Agreed, let's merge this with Sylvain's suggestions and make it easy for folks to contribute.

PR #55, opened 2023-06-12
Hide semaphore technical implementation, and use an interface

dmathieu:
Exposing the `Semaphore` config is exposing an implementation detail. Services using this library only care about the effect of the semaphore (the maximum number of concurrencies), not that it's using a channel.

Also, by exposing a channel, we can't change the underlying implementation, and (for example) switch to [golang.org/x/sync/semaphore](https://pkg.go.dev/golang.org/x/sync/semaphore).

### DIFF
--- a/input/semaphore.go
+++ b/input/semaphore.go
@@ -21,7 +21,14 @@ import "context"
 
 // Semaphore is used by inputs to limit the number of concurrent requests
 type Semaphore interface {
+	// Acquire acquires the semaphore with a weight of n, blocking until
+	// resources are available or ctx is done. On success, returns nil. On
+	// failure, returns ctx.Err() and leaves the semaphore unchanged.
 	Acquire(context.Context, int64) error
+	// TryAcquire acquires the semaphore with a weight of n without blocking. On
+	// success, returns true. On failure, returns false and leaves the semaphore
+	// unchanged.
 	TryAcquire(int64) bool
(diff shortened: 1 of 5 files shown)


marclop (commented):
I also prefer this interface, however, with this approach, we can't share the same limit between more than one input (say `OTLP` and `elasticapm`). Ideally, we'd create the concurrency semaphore outside of the processors, pass it to the relevant inputs, and this way, we'll have the same pool of concurrency for all inputs.

Something that we may consider is creating an interface:

```go
type Semaphore interface {
	Acquire(context.Context, int) error
	Release(int)
}
```

And passing in the type that satisfies it.

dmathieu:
I see your point about sharing the semaphore, and using an interface is definitely better. I've made that change.
That makes it impossible not to do a breaking change though.

input/semaphore.go
```diff
@@ -0,0 +1,27 @@
...
+
+package input
+
+import "context"
+
+// Semaphore is used by inputs to limit the number of concurrent requests
+type Semaphore interface {
+	Acquire(context.Context, int64) error
+	TryAcquire(int64) bool
+	Release(int64)
```
marclop, line 26: Small nit, can we add the expectations for implementers? What should each method do and how to behave. Copying the current `sync/semaphore.Weighted` docs is probably good enough.

marclop (approved):
Very clean ✅

PR #56, opened 2023-06-12
Setup otlp max concurrency semaphore

dmathieu:
This adds a `MaxConcurrency` config for the OTLP consumer, and sets up a semaphore to limit the maximum number of concurrent exports.

See https://github.com/elastic/apm-server/issues/9242
Related: https://github.com/elastic/apm-data/pull/55


marclop (approved):
LGTM!

PR #57, opened 2023-06-12
feat: add protobuf utils and improve compatibility for easier migration

kruskall:
This PR is adding a few methods for easier migration to protobuf in APM Server:

- `ToModelProtobuf` is made public to allow converting from classic/old `model` to protobuf `modelpb`
- Add a `ProtoBatchProcessor` as a compatibility layer between `model.BatchProcessor` and `modelpb.BatchProcessor`
- Add `modelpb.BatchProcessor`: required for migrating to protobuf
- `modelpb` labels: these are type aliases with utility methods, similar to what we have for the classic/old `model`s
- url utility methods similar to what we had for the classic/old `model`s

### DIFF
--- a/model/apmevent.go
+++ b/model/apmevent.go
@@ -91,11 +91,11 @@ func (e *APMEvent) MarshalJSON() ([]byte, error) {
 // MarshalFastJSON marshals e as JSON, writing the result to w.
 func (e *APMEvent) MarshalFastJSON(w *fastjson.Writer) error {
 	var outProto modelpb.APMEvent
-	e.toModelProtobuf(&outProto)
+	e.ToModelProtobuf(&outProto)
 	return outProto.MarshalFastJSON(w)
 }
 
-func (e *APMEvent) toModelProtobuf(out *modelpb.APMEvent) {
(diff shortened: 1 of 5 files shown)


model/apmevent.go
```diff
@@ -91,11 +91,11 @@ func (e *APMEvent) MarshalJSON() ([]byte, error) {
 // MarshalFastJSON marshals e as JSON, writing the result to w.
 func (e *APMEvent) MarshalFastJSON(w *fastjson.Writer) error {
 	var outProto modelpb.APMEvent
-	e.toModelProtobuf(&outProto)
+	e.ToModelProtobuf(&outProto)
 	return outProto.MarshalFastJSON(w)
 }
 
-func (e *APMEvent) toModelProtobuf(out *modelpb.APMEvent) {
+func (e *APMEvent) ToModelProtobuf(out *modelpb.APMEvent) {
```
dmathieu, line 98: Shouldn't there be a comment? (I'm a bit surprised the linter doesn't require one)

kruskall: I guess we could add it. Keep in mind, this is temporary and it's public just to make it easier to migrate. The `model` package will go away along with this method.

dmathieu (approved):

Issue #52, opened 2023-06-08
Integrate protobuf definitions for model types in apm-server


simitt:
Follow up on https://github.com/elastic/apm-data/issues/36

```[tasklist]
## Phase 2
- [x] Update apm-server to use protoc-generated Go types everywhere, and removing handwritten model types
- [x] Update apm-data to decode events directly into protobuf-generated types, and remove the old model types
- [x] Update apm-server to encode events to protobuf for tail-based storage
```

kruskall:
For the testplan or whoever ends up testing this, here are a few things to test/look out for:

- models are no longer using `map[string]any` directly, instead the data from APM Agents is mapped to `structpb.Struct` and then converted to `map[string]any` in modeljson. There is an additional point of failure in this approach because the creation of `structpb.Struct` could fail. If that happens, we lose data.
- protobuf models are using pointers everywhere. Look out for nil pointer dereference. `event.Foo.Bar` was safe before but now it's not.
- IPs are stored as strings as opposed to `netip.Addr`. Make sure we are not converting an invalid IP (`invalidNetip.String()` will produce "invalid ip", we probably shouldn't encode that).
- protobus models are passing by reference (they are not safe to copy), if we need a copy we should use `proto.Clone`

kruskall:
The listed tasks have been completed. Closing this.

Issue #47, opened 2023-06-06
Review protobuf types before declaring it stable


kruskall:
We should review protobuf fields to ensure we are using proper types.

- [x] validate signed/unsigned int usage
- [x] make sure ints have proper size (32/64)
- [x] improved storage for IPs (followup from https://github.com/elastic/apm-data/pull/42#discussion_r1213841328)

axw:
We should also look at the encoding/decoding performance of maps, and compare against the array-of-structs approach: https://protobuf.dev/programming-guides/proto3/#backwards. We may want to do that after integrating into apm-server, in case it makes operating on the model types too painful.

kruskall:
Also address `processor` field improvement: https://github.com/elastic/apm-data/pull/58#discussion_r1228975554

axw:
I think we should probably not use the `Struct` type everywhere we're using it at the moment. e.g. HTTP headers should probably be `repeated Header`, where Header is a message comprising `string Key` and `repeated string Values`. Similar to Labels, but without Global.

axw:
[SpanLink](https://github.com/elastic/apm-data/blob/5cae1b218218f79f28299d3201c7024f893105f1/model/proto/span.proto#L75C19-L78) should be updated to be a pair of string fields.

axw:
Parent should be replaced with a ParentID field on the relevant types. Wherever we're using Span, Transaction, and Trace just to set the ID field (like in SpanLink) we should replace them with SpanID, TransactionID, TraceID.

axw:
Same for Child, should be replaced with `repeated string`

axw:
We should investigate whether we could use the Elasticsearch [`uri_parts`](https://www.elastic.co/guide/en/elasticsearch/reference/current/uri-parts-processor.html) ingest processor to parse URLs, and include only the full URL in the model types.

dmathieu:
> [SpanLink](https://github.com/elastic/apm-data/blob/5cae1b218218f79f28299d3201c7024f893105f1/model/proto/span.proto#L75C19-L78) should be updated to be a pair of string fields.

I'm doubting a bit about that. OTLP has more to links that just a trace/span ID. Notably, it can have attributes.
https://github.com/open-telemetry/opentelemetry-proto/blob/c4dfbc51f3cd4089778555a2ac5d9bc093ed2956/opentelemetry/proto/trace/v1/trace.proto#L220-L239

axw:
[...]

PR #58, opened 2023-06-12
feat: migrate batch processing to protobuf models

kruskall:
Migrate BatchProcessor to protobuf models 

Blocked by https://github.com/elastic/apm-data/pull/57
Blocked by https://github.com/elastic/apm-data/pull/59

### DIFF
--- a/model/modelprocessor/chained.go
+++ b/model/modelprocessor/chained.go
@@ -21,13 +21,25 @@ import (
 	"context"
 
 	"github.com/elastic/apm-data/model"
+	"github.com/elastic/apm-data/model/modelpb"
 )
 
 // Chained is a chained model.BatchProcessor, calling each of
 // the processors in the slice in series.
-type Chained []model.BatchProcessor
+type PbChained []modelpb.BatchProcessor
 
 // ProcessBatch calls each of the processors in c in series.
+func (c PbChained) ProcessBatch(ctx context.Context, batch *modelpb.Batch) error {
+	for _, p := range c {
+		if err := p.ProcessBatch(ctx, batch); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+type Chained []model.BatchProcessor
+
 func (c Chained) ProcessBatch(ctx context.Context, batch *model.Batch) error {
 	for _, p := range c {
 		if err := p.ProcessBatch(ctx, batch); err != nil {
--- a/model/modelprocessor/chained.go
+++ b/model/modelprocessor/chained.go
@@ -24,7 +24,7 @@ import (
 	"github.com/elastic/apm-data/model/modelpb"
 )
 
-// Chained is a chained model.BatchProcessor, calling each of
+// PbChained is a chained modelpb.BatchProcessor, calling each of
 // the processors in the slice in series.
 type PbChained []modelpb.BatchProcessor
 
@@ -38,8 +38,11 @@ func (c PbChained) ProcessBatch(ctx context.Context, batch *modelpb.Batch) error
 	return nil
 }
 
+// Chained is a chained model.BatchProcessor, calling each of
+// the processors in the slice in series.
 type Chained []model.BatchProcessor
 
+// ProcessBatch calls each of the processors in c in series.
 func (c Chained) ProcessBatch(ctx context.Context, batch *model.Batch) error {
 	for _, p := range c {
 		if err := p.ProcessBatch(ctx, batch); err != nil {
--- a/model/modelprocessor/datastream.go
+++ b/model/modelprocessor/datastream.go
@@ -22,7 +22,7 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/elastic/apm-data/model"
+	"github.com/elastic/apm-data/model/modelpb"
 )
 
 const (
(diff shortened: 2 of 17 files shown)


kruskall:
Tests should pass once https://github.com/elastic/apm-data/pull/59 is merged

model/modelprocessor/hostname.go
```diff
@@ -29,25 +29,25 @@ import (
...
-	case event.Kubernetes.PodName != "" || event.Kubernetes.PodUID != "" || event.Kubernetes.Namespace != "":
+	case event.GetKubernetes().GetPodName() != "" || event.GetKubernetes().GetPodUid() != "" || event.GetKubernetes().GetNamespace() != "":
 		// kubernetes.* is set, but kubernetes.node.name is not: don't set host.hostname at all.
 		event.Host.Hostname = ""
 	default:
 		// Otherwise use the originally specified host.hostname value.
 	}
-	if event.Host.Name == "" {
+	if event.GetHost().GetName() == "" && event.GetHost().GetHostname() != "" {
 		event.Host.Name = event.Host.Hostname
```
marclop, line 51: We're setting the hostname here and above without checking if `event.Host != nil`, something like this is safer:

```go
func setHostHostname(event *modelpb.APMEvent) {
	k8s := event.GetKubernetes()
	if event.GetHost() == nil {
		event.Host = &modelpb.Host{}
	}
	switch {
	case k8s.GetNodeName() != "":
		// host.kubernetes.node.name is set: set host.hostname to its value.
		event.Host.Hostname = k8s.NodeName
	case k8s.GetPodName() != "" || k8s.GetPodUid() != "" || k8s.GetNamespace() != "":
		// kubernetes.* is set, but kubernetes.node.name is not: don't set host.hostname at all.
		event.Host.Hostname = ""
	default:
		// Otherwise use the originally specified host.hostname value.
	}
	if host := event.GetHost(); host.GetName() == "" && host.GetHostname() != "" {
		event.Host.Name = event.Host.Hostname
	}
}
```

model/modelprocessor/chained.go
```diff
@@ -21,13 +21,25 @@ import (
 	"context"
 
 	"github.com/elastic/apm-data/model"
+	"github.com/elastic/apm-data/model/modelpb"
 )
 
 // Chained is a chained model.BatchProcessor, calling each of
```
axw, line 27: Comment needs updating.

model/modelprocessor/errormessage.go
```diff
@@ -20,25 +20,25 @@ package modelprocessor
...
+		event := (*b)[i]
 		if event.Error != nil {
 			event.Message = s.setErrorMessage(event)
 		}
 	}
 	return nil
 }
 
-func (s SetErrorMessage) setErrorMessage(event *model.APMEvent) string {
+func (s SetErrorMessage) setErrorMessage(event *modelpb.APMEvent) string {
```
marclop, line 41: This function can be simplified using getters:

```go
func (s SetErrorMessage) setErrorMessage(event *modelpb.APMEvent) string {
	if msg := event.GetError().GetLog().GetMessage(); msg != "" {
		return msg
	}
	if msg := event.GetError().GetException().GetMessage(); msg != "" {
		return msg
	}
	return ""
}
```

model/modelprocessor/dropunsampled.go
```diff
@@ -51,9 +51,9 @@ func NewDropUnsampled(dropRUM bool, droppedCallback func(int64)) model.BatchProc
 	})
 }
 
-func shouldDropUnsampled(event *model.APMEvent, dropRUM bool) bool {
-	if event.Processor != model.TransactionProcessor || event.Transaction == nil || event.Transaction.Sampled {
+func shouldDropUnsampled(event *modelpb.APMEvent, dropRUM bool) bool {
+	if !event.Processor.IsTransaction() || event.Transaction == nil || event.Transaction.Sampled {
```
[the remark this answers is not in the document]
kruskall: It would but `IsTransaction` is already using getters under the hood so this should panic even if `event.Processor` is nil.

Is this only for clarity/style or am I missing something ?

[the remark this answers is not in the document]
marclop: Not because it is unsafe, but I figured that 2 `||` clauses seem easier to read than 3.

model/modelprocessor/hostname.go
kruskall: In this case `event.GetHost().GetHostname() != ""` guarantees that `event.Host` is not nil 

For the switch case above: I agree, good catch. We still need to make sure to only set the host if the hostname will actually be set

model/modelprocessor/errormessage.go
kruskall: Good catch!

model/modelprocessor/dropunsampled.go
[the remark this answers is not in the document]
kruskall: Oh sorry, I stopped reading at `Processor` -> `GetProcessor` :facepalm: 

I didn't notice the transaction change, I agree it looks cleaner :+1:

model/modelprocessor/datastream.go
```diff
@@ -40,40 +40,43 @@ const (
...
 	}
 	return nil
 }
 
-func (s *SetDataStream) setDataStream(event *model.APMEvent) {
-	switch event.Processor {
-	case model.SpanProcessor, model.TransactionProcessor:
+func (s *SetDataStream) setDataStream(event *modelpb.APMEvent) {
+	switch {
+	case event.GetProcessor().IsSpan(), event.GetProcessor().IsTransaction():
```
axw, line 65: The "processor" fields are a bit of a relic, and we should aim to remove them in the long term. With that in mind, I wonder if we should change the model a little bit so we can remove them from the apm-data codebase, and set the `processor.*` fields in our ingest pipelines, or by setting a value on `constant_keyword` fields where it makes sense.

e.g. for metrics, `processor.name` and `processor.event` are both always "metric", so we can update their field definitions to set the value in the mapping: https://github.com/elastic/apm-server/blob/23fb1577909836ebf45e65705df3fd560de5adb1/apmpackage/apm/data_stream/app_metrics/fields/fields.yml#L30-L35

IIANM the only exception to this is for the `apm.traces` and `apm.rum` data streams, where spans and transactions end up. These both have have `processor.name: transaction`, but they differ in `processor.event` (one is "span", one is "transaction"). Eventually spans and transactions should converge, but for now I think we could set the value in an ingest pipeline.

Maybe we could:
 - set [`event.kind`](https://www.elastic.co/guide/en/ecs/current/ecs-event.html#field-event-kind) to either "span" or "transaction" in the apm-data code
 - update the `traces` ingest pipeline to use this to populate `processor.event`, and then remove `event.kind` since those values are not valid for `event.kind`

WDYT?

axw: Doesn't have to be done right now btw, but I think we should do it as part of https://github.com/elastic/apm-data/issues/47

model/modelprocessor/chained.go
kruskall: Fixed :+1:

model/modelprocessor/datastream.go
kruskall: > Doesn't have to be done right now btw, but I think we should do it as part of #47

kruskall: > The "processor" fields are a bit of a relic, and we should aim to remove them in the long term. With that in mind, I wonder if we should change the model a little bit so we can remove them from the apm-data codebase, and set the `processor.*` fields in our ingest pipelines, or by setting a value on `constant_keyword` fields where it makes sense.
> 
> e.g. for metrics, `processor.name` and `processor.event` are both always "metric", so we can update their field definitions to set the value in the mapping: https://github.com/elastic/apm-server/blob/23fb1577909836ebf45e65705df3fd560de5adb1/apmpackage/apm/data_stream/app_metrics/fields/fields.yml#L30-L35
> 
> IIANM the only exception to this is for the `apm.traces` and `apm.rum` data streams, where spans and transactions end up. These both have have `processor.name: transaction`, but they differ in `processor.event` (one is "span", one is "transaction"). Eventually spans and transactions should converge, but for now I think we could set the value in an ingest pipeline.
> 
> Maybe we could:
> 
>     * set [`event.kind`](https://www.elastic.co/guide/en/ecs/current/ecs-event.html#field-event-kind) to either "span" or "transaction" in the apm-data code
> 
>     * update the `traces` ingest pipeline to use this to populate `processor.event`, and then remove `event.kind` since those values are not valid for `event.kind`

I think it's a good plan, we would probably need to coordinate with the kibana team to remove them in the long term to make sure we aren't breaking anything they are relying on

axw (approved):
LGTM, with the exception of the inresolved conversation about setting the hostname. Please wait for @marclop.

model/modelprocessor/hostname.go
kruskall: Forgot to leave a comment: I've updated the code

model/modelprocessor/dropunsampled.go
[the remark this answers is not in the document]
kruskall: Updated!

model/modelprocessor/hostname.go
marclop: It doesn't look right, the `event.Host` is being set inside the switch case, which may or may not be set. In cases where the Kubernetes and event.Host structs are nil, line 56 will panic.

kruskall: I think I'm missing something. Line 56 is guarded by:

`if event.GetHost().GetName() == "" && event.GetHost().GetHostname() != "" {`

The only way for `event.GetHost().GetHostname() != ""` to be true is if `event.Host` is not nil so it shouldn't be possible for line 56 to panic.

marclop: Nothing is missing. I misread the second part of the if clause.

marclop (approved):
LGTM

PR #60, opened 2023-06-13
feat: migrate otlp to protobuf models 

kruskall:
migrate otlp to protobuf models

Blocked by https://github.com/elastic/apm-data/pull/58

### DIFF
--- a/input/elasticapm/internal/modeldecoder/v2/decoder.go
+++ b/input/elasticapm/internal/modeldecoder/v2/decoder.go
@@ -1430,7 +1430,7 @@ func mapOTelAttributesTransaction(from otel, out *model.APMEvent) {
 	// potentially in the actual attributes map?
 	spanStatus := ptrace.NewStatus()
 	spanStatus.SetCode(ptrace.StatusCodeUnset)
-	otlp.TranslateTransaction(m, spanStatus, scope, out)
+	otlp.TranslateTransactionOld(m, spanStatus, scope, out)
 
 	if out.Span.Kind == "" {
 		switch out.Transaction.Type {
@@ -1478,7 +1478,7 @@ func mapOTelAttributesSpan(from otel, out *model.APMEvent) {
 		}
 		out.Span.Kind = from.SpanKind.Val
 	}
-	otlp.TranslateSpan(spanKind, m, out)
+	otlp.TranslateSpanOld(spanKind, m, out)
 
 	if spanKind == ptrace.SpanKindUnspecified {
 		switch out.Span.Type {
--- a/input/otlp/metadata.go
+++ b/input/otlp/metadata.go
@@ -44,238 +44,144 @@ func translateResourceMetadata(resource pcommon.Resource, out *modelpb.APMEvent)
 		switch k {
 		// service.*
 		case semconv.AttributeServiceName:
-			if out.Service == nil {
-				out.Service = &modelpb.Service{}
-			}
+			out.Service = populateNil(out.Service)
 			out.Service.Name = cleanServiceName(v.Str())
 		case semconv.AttributeServiceVersion:
-			if out.Service == nil {
-				out.Service = &modelpb.Service{}
-			}
+			out.Service = populateNil(out.Service)
 			out.Service.Version = truncate(v.Str())
 		case semconv.AttributeServiceInstanceID:
 			out.Service.Node.Name = truncate(v.Str())
 
 		// deployment.*
 		case semconv.AttributeDeploymentEnvironment:
-			if out.Service == nil {
-				out.Service = &modelpb.Service{}
-			}
+			out.Service = populateNil(out.Service)
 			out.Service.Environment = truncate(v.Str())
 
 		// telemetry.sdk.*
 		case semconv.AttributeTelemetrySDKName:
-			if out.Agent == nil {
-				out.Agent = &modelpb.Agent{}
-			}
+			out.Agent = populateNil(out.Agent)
 			out.Agent.Name = truncate(v.Str())
 		case semconv.AttributeTelemetrySDKVersion:
-			if out.Agent == nil {
-				out.Agent = &modelpb.Agent{}
-			}
+			out.Agent = populateNil(out.Agent)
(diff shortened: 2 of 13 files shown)


input/otlp/traces.go
```diff
@@ -365,39 +436,43 @@ func TranslateTransaction(
 
 	switch foundSpanType {
 	case httpSpan:
-		event.HTTP = http
+		if !proto.Equal(&http, &modelpb.HTTP{}) {
+			event.Http = &http
+		}
```
axw, line 441: If `foundSpanType == httpSpan`, shouldn't we know for a fact that http is non-empty?

input/otlp/traces.go
```diff
@@ -670,14 +791,19 @@ func TranslateSpan(spanKind ptrace.SpanKind, attributes pcommon.Map, event *mode
...
 			}
 		}
 		event.Span.Type = "external"
 		subtype := "http"
 		event.Span.Subtype = subtype
-		event.HTTP = http
-		event.URL.Original = httpURL
+		if !proto.Equal(&http, &modelpb.HTTP{}) {
+			event.Http = &http
+		}
```
axw, line 802: same as above, shouldn't it be expected to be non-empty? can we drop the proto.Equal?

input/otlp/traces_old.go
```diff
@@ -0,0 +1,665 @@
+package otlp
```
axw, line 1: Missing copyright

input/otlp/traces_old.go
```diff
@@ -0,0 +1,665 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/elastic/apm-data/model"
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+	semconv "go.opentelemetry.io/collector/semconv/v1.5.0"
+	"google.golang.org/grpc/codes"
+)
+
```
axw, line 17: Maybe add a warning comment here and in traces.go that any changes to these need to be kept in sync?

axw (commented):
Mostly LGTM, just a few small things

input/otlp/traces_old.go
kruskall: Updated! :+1:

input/otlp/traces.go
kruskall: Removed :+1:

input/otlp/traces.go
kruskall: Removed :+1:

input/otlp/metadata.go
```diff
@@ -163,62 +303,86 @@ func translateResourceMetadata(resource pcommon.Resource, out *model.APMEvent) {
...
+			}
+			if out.Service.Language == nil {
+				out.Service.Language = &modelpb.Language{}
+			}
 			out.Service.Language.Name = versionParts[1]
 		}
 		if v := versionParts[len(versionParts)-1]; v != "" {
+			if out.Agent == nil {
+				out.Agent = &modelpb.Agent{}
+			}
```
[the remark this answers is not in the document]
kruskall: Updated :+1:

input/otlp/traces.go
kruskall: It seems that's not the case. See the failing test:

```
 === RUN   TestConsumer_JaegerSpan/jaeger_https_default_port
    traces_test.go:1224:   map[string]any{
          	"events": []any{
          		map[string]any{
          			... // 3 identical entries
          			"event":     map[string]any{"duration": float64(7.9e+10), "outcome": string("unknown")},
          			"host":      map[string]any{"hostname": string("host-abc")},
        + 			"http":      map[string]any{},
          			"parent":    map[string]any{"id": string("0000000058585858")},
          			"processor": map[string]any{"event": string("span"), "name": string("transaction")},
          			... // 5 identical entries
          		},
          	},
          }
```

axw: Hmm true, because you might have just `http.url`, which translates to `url.*` fields in our model. Let's revisit this later to see if we can find a way to avoid this `proto.Equal` call.

axw (approved):

input/otlp/metadata.go
```diff
@@ -38,104 +38,244 @@ var (
...
 		// service.*
 		case semconv.AttributeServiceName:
+			if out.Service == nil {
+				out.Service = &modelpb.Service{}
+			}
 			out.Service.Name = cleanServiceName(v.Str())
 		case semconv.AttributeServiceVersion:
+			if out.Service == nil {
+				out.Service = &modelpb.Service{}
+			}
```
marclop, line 54: All of this `out.<Field> == nil`  seems a bit verbose. Would it make sense to introduce a generic function that checks if the field is nil, and if it is, it allocates the new type? Something like this would work:

```go
func translateResourceMetadata(resource pcommon.Resource, out *modelpb.APMEvent) {
	[...]
	out.Service = populateNil(out.Service)
	out.Service.Version = truncate(v.Str())
	[...]
}

func populateNil[T any](a *T) *T {
	if a == nil {
		var t T
		return &t
	}
	return a
}
```

marclop (commented):
Looks good, just a nit on the nil checking.

input/elasticapm/internal/modeldecoder/v2/decoder.go
```diff
@@ -1430,7 +1430,7 @@ func mapOTelAttributesTransaction(from otel, out *model.APMEvent) {
 	// potentially in the actual attributes map?
 	spanStatus := ptrace.NewStatus()
 	spanStatus.SetCode(ptrace.StatusCodeUnset)
-	otlp.TranslateTransaction(m, spanStatus, scope, out)
+	otlp.TranslateTransactionOld(m, spanStatus, scope, out)
```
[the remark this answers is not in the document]
kruskall: This is part of the decoder work so we need to use the old `model`

input/otlp/metadata.go
kruskall: Thanks a lot for this! :bow:

marclop (approved):

PR #61, opened 2023-06-13
fix: only map selftime.sum to protobuf if the sum is not zero

kruskall:
Small fix for protobuf mapping.

sum is a pointer type, we should only set one if the value is not empty to have the same behaviour as the old/classic models


dmathieu (approved):
Should there be tests?

kruskall:
> Should there be tests?

This is indirectly tested by the additional commit migrating to protobuf, I don't think we should spend more effort in adding more coverage than we already are as this file/code is going away once we fully switch to protobuf.

PR #64, opened 2023-06-15
codec: Add codec abstraction, JSON implementation

marclop:
Adds a new top level package `codec` which defines the interfaces to implement a codec, and two helper functions to wrap an encoder / decoder with encoding or decoding metrics. Also adds a new `codec/json` package which implementing a JSON codec over the standard `encoding/json` pkg.


codec/codec.go
```diff
@@ -0,0 +1,81 @@
...
+type Encoder interface {
+	// Encode encodes a type into its byte slice representation.
+	Encode(in any) ([]byte, error)
+}
+
+// Decoder decodes a byte slice representation into its Go type.
+type Decoder interface {
+	// Decode decodes a byte slice representation into its Go type.
+	Decode(in []byte, out any) error
+}
```
axw, line 38: Should we use https://pkg.go.dev/encoding#BinaryMarshaler and https://pkg.go.dev/encoding#BinaryUnmarshaler?

marclop: I was self commenting the code to bring up whether or not we should use existing stdlib interfaces

marclop: I was looking into those and while they're similar, they don't exactly adhere to the re-usable and concurrently safe `Codec` introduced here, `MarshalBinary` => `Encode`, `UnmarshalBinary` => `Decode`, but the issue doesn't lay with the naming but rather, the defined methods do not accept an `in` (for `MarshalBinary`) or `out` (for `UnmarshalBinary`:

```go
type combined interface {
	MarshalBinary() (data []byte, err error)
	UnmarshalBinary(data []byte) error
}
```

This is a problem if the codec is to be used concurrently since it means that we'd have to have an individual codec per goroutine and sharing isn't possible:

```go
var a map[string]string{"a":"b"}
codec := json.Codec{Data: a}
b, err := codec.MarshalBinary()
```

We could take the same approach the `encoding/json` library takes and use pooled encoders: 

```go
func Marshal(v any) ([]byte, error) {
	e := newEncodeState()
	defer encodeStatePool.Put(e)

	err := e.marshal(v, encOpts{escapeHTML: true})
	if err != nil {
		return nil, err
	}
	buf := append([]byte(nil), e.Bytes()...)

	return buf, nil
}

var encodeStatePool sync.Pool

func newEncodeState() *encodeState {
	if v := encodeStatePool.Get(); v != nil {
		e := v.(*encodeState)
		e.Reset()
		if len(e.ptrSeen) > 0 {
			panic("ptrEncoder.encode should have emptied ptrSeen via defers")
		}
		e.ptrLevel = 0
		return e
	}
	return &encodeState{ptrSeen: make(map[any]struct{})}
}
```

For the protoc-generated models with the Vitess plugin, the names are `MarshalVT` and `UnmarshalVT`:

https://github.com/elastic/apm-data/blob/a0d42115e6b7eead3c4b018b9ad9aa842e0d5492/model/modelpb/agent_vtproto.pb.go#L39-L50

https://github.com/elastic/apm-data/blob/a0d42115e6b7eead3c4b018b9ad9aa842e0d5492/model/modelpb/agent_vtproto.pb.go#L143

Which match our interfaces exactly, we'd just need to create an abstraction that wraps the calls to those methods with the names that we define in `codec.Codec`:

```go
type Marshaler interface {
	MarshalVT() (dAtA []byte, err error)
}

type Unmarshaler interface {
	UnmarshalVT(dAtA []byte) error
}

type Codec struct{}

// Encode encodes a type into the JSON byte slice representation.
func (c Codec) Encode(in any) ([]byte, error) {
	if m, ok := in.(Marshaler); ok {
		return m.MarshalVT()
	}
	return nil, errors.New("protobuf: invalid type; not compliant with Marshaler")
}

// Decode decodes a JSON-encoded byte slice into its type.
func (c Codec) Decode(in []byte, out any) error {
	if m, ok := out.(Unmarshaler); ok {
		return m.UnmarshalVT(in)
	}
	return errors.New("protobuf: invalid type; not compliant with Unmarshaler")
}
```

axw: > I was looking into those and while they're similar, they don't exactly adhere to the re-usable and concurrently safe Codec introduced here, MarshalBinary => Encode, UnmarshalBinary => Decode, but the issue doesn't lay with the naming but rather, the defined methods do not accept an in (for MarshalBinary) or out (for UnmarshalBinary:

Right, of course :man_facepalming:

As discussed live, I think what you have is fine and we can take another look at it when we are using the new approach in users of apm-data/apm-queue.

axw (approved):

codec/codec.go
```diff
@@ -0,0 +1,81 @@
...
+
+// Codec is a composite of Encoder and Decoder.
+type Codec interface {
+	Encoder
+	Decoder
+}
+
+// RecordEncodedBytes decorates an encoder with a metric that records the bytes
+// bytes that have been encoded.
```
kyungeunni, line 47:
```suggestion
// that have been encoded.
```

kyungeunni (approved):
LGTM!

PR #69, opened 2023-06-19
perf: bump fastjson for string encoding improvements

kruskall:
Bump fastjson so that we can use https://github.com/elastic/go-fastjson/pull/16

This should improve modeljson marshaling performance


go.mod
```diff
@@ -12,15 +12,15 @@ require (
 	github.com/stretchr/testify v1.8.3
 	github.com/xeipuuv/gojsonschema v1.2.0
 	go.elastic.co/apm/v2 v2.2.0
-	go.elastic.co/fastjson v1.1.0
+	go.elastic.co/fastjson v1.2.1-0.20230619075946-485c9d6890d9
```
axw, line 15: Can we create a new tag in go-fastjson, rather than pinning to a commit?

kruskall: Sounds good to me! Should we release v1.3.0 ? v1.2.1 ?

axw: I'd say v1.3.0, since it's not a bug fix.

kruskall: Should be fixed!

axw (approved):
thanks :)

PR #75, opened 2023-06-19
refactor: standardize headers type

kruskall:
Related to https://github.com/elastic/apm-data/issues/47

Should solve https://github.com/elastic/apm-data/issues/47#issuecomment-1595599688

Also avoid cloning the map since we're only reading from it and copying the values to a new struct.

Is there a reason why we are not going with `map[string]*HeaderValue` ? I don't feel strongly about it, just curious why we went with a slice.


input/elasticapm/internal/modeldecoder/modeldecoderutil/http.go
```diff
@@ -45,17 +45,18 @@ func HTTPHeadersToStructPb(h http.Header) *structpb.Struct {
...
 	}
-	m := make(map[string]*modelpb.HTTPHeaderValue, len(h))
+	headers := make([]*modelpb.HTTPHeader, 0, len(h))
 	for k, v := range h {
-		m[k] = &modelpb.HTTPHeaderValue{
-			Values: v,
-		}
+		headers = append(headers, &modelpb.HTTPHeader{
+			Key:   k,
+			Value: v,
```
axw, line 56: Should we be copying the slice here? Previously we were cloning the `http.Header` value when translating, so any mutation of the slice wouldn't impact the original value. I think this matters since the decoder reuses memory.

axw (approved):
LGTM but I think we need to copy the header value slice in HTTPHeadersToModelpb.

> Is there a reason why we are not going with map[string]*HeaderValue ? I don't feel strongly about it, just curious why we went with a slice.

I suppose it doesn't matter much. The wire protocol is the same. We don't need to do any lookups on the headers though, so it seems unnecessary; a slice will be more compact and efficient for basic copy/append/iterate operations.

input/elasticapm/internal/modeldecoder/modeldecoderutil/http.go
kruskall: I might be missing something but isn't the object reset before putting it back into the pool ? The original `http.Header` is cleared and the slice never mutated. When could it be mutated ?

axw: Any model processor _could_ modify the slice. They might not now, but we need to be defensive here since the input model is pooled and could be reused across two different connections.

kruskall: Sorry, I didn’t understand. Can you clarify:

Suppose we are mapping the `errorEvent`, the input model is fetched from the pool, mapped to modelpb and released into the pool: https://github.com/elastic/apm-data/blob/c6848e06bed4f3cd116ed203b13bcf4e4cd3887b/input/elasticapm/internal/modeldecoder/v2/decoder.go#L173-L174

Before being released, the object is reset: https://github.com/elastic/apm-data/blob/c6848e06bed4f3cd116ed203b13bcf4e4cd3887b/input/elasticapm/internal/modeldecoder/v2/decoder.go#L105

The reset call is recursive and it's gonna clear the `HTTPHeader`: https://github.com/elastic/apm-data/blob/c6848e06bed4f3cd116ed203b13bcf4e4cd3887b/input/elasticapm/internal/modeldecoder/nullable/nullable.go#L316

My understanding is that the original `http.Header` is cleared and all the slices it contains are never mutated. Once the original object is reset and the http header is cleared, the only reference to the "original values" (the slices) is the one we passed to the `modelpb`.

The model processors would then be able to modify the `modelpb` events which contains slices that are no longer referenced by the input models.

axw: I was thinking that the `http.Header` was pooled as well, but it's not - a new one is created in the nullable.HTTPHeader json-iterator callback. So this should be safe after all.

axw (approved):

PR #79, opened 2023-06-20
Remove APMEvent.Processor, infer type

axw:
Remove the APMEvent.Processor type, which is a relic of the past when APM Server had different event processors and we set this field according to which processor created the document.

Instead of setting these fields explicitly, we can infer the type of an event from other fields:
 - if Metricset is non-nil, it's a metric
 - if Error is non-nil, it's an error
 - if Log is non-nil, it's a log
 - if Transaction.Type is set, it's a transaction
 - if Span.Type is set, it's a span

One additional rule, since there may be logs without `log.*` fields (hence Log will be nil): if `event.kind` is "event", then we treat that as a log as well.

We still use `processor.*` fields in various places in the UI. Until/unless we remove them from there as well, we can infer them during ingest, or set them as constant_keyword values in the mapping based on the datastream.

Companion PR in apm-server: https://github.com/elastic/apm-server/pull/11028


input/elasticapm/internal/modeldecoder/v2/decoder.go
```diff
@@ -1549,6 +1543,10 @@ func mapToLogModel(from *log, event *modelpb.APMEvent) {
 	if len(from.Labels) > 0 {
 		modeldecoderutil.MergeLabels(from.Labels, event)
 	}
+	if event.Error == nil {
+		event.Event = populateNil(event.Event)
+		event.Event.Kind = "event"
+	}
```
kruskall, line 1549: **question**: was this lost during the protobuf migration ? Is this something that we should backport to 8.9 ?

input/otlp/traces.go
```diff
@@ -924,10 +922,11 @@ func (c *Consumer) convertSpanEvent(
 	}
 
 	if event.Error != nil {
-		event.Processor = modelpb.ErrorProcessor()
 		setErrorContext(event, parent)
 	} else {
-		event.Processor = modelpb.LogProcessor()
+		// Set "event.kind" to indicate this is a log event.
+		event.Event = populateNil(event.Event)
+		event.Event.Kind = "event"
```
kruskall, line 929: same as above

kruskall (approved):
LGTM! Left a minor question

input/elasticapm/internal/modeldecoder/v2/decoder.go
axw: No, I added this as a means to identify a document as being a log prior to its `data_stream.*` fields being set. No need to backport.

input/otlp/traces.go
axw: same :)

axw:
https://github.com/elastic/apm-server/pull/11202 needs to be merged and bundled with Kibana before this is used by apm-server, to avoid breaking everything.

marclop (approved):
Looks great! Less to encode and send over the wire 👍🏼

PR #84, opened 2023-06-23
feat: add vtproto codec

kyungeunni:
part of https://github.com/elastic/apm-managed-service/issues/3

Use `MarshalVT()` and `UnmarshalVT()` to encode and decode.


kruskall (approved):

codec/vtproto/codec.go
```diff
@@ -0,0 +1,47 @@
...
+	MarshalVT() ([]byte, error)
+	UnmarshalVT([]byte) error
+}
+
+// Codec is a composite of Encoder and Decoder.
+type Codec struct {
+}
+
+// Encode encodes vtprotoMessage type into byte slice
+func (Codec) Encode(in any) ([]byte, error) {
```
marclop, line 32: rather than accept `in any` and then doing a type assertion on top, I think it would be better to the compiler tell us when the passed type isn't right, rather than throw a runtime error.

codec/vtproto/codec.go
```diff
@@ -0,0 +1,47 @@
...
+	vt, ok := in.(vtprotoMessage)
+	if !ok {
+		return nil, fmt.Errorf("failed to encode, message is %T (missing vtprotobuf helpers)", in)
+	}
+	return vt.MarshalVT()
+}
+
+// Decode decodes a byte slice into vtprotoMessage type.
+func (Codec) Decode(in []byte, out any) error {
```
marclop, line 41: same here.

codec/vtproto/codec.go
```diff
@@ -0,0 +1,47 @@
...
+// under the License.
+
+package vtproto
+
+import "fmt"
+
+type vtprotoMessage interface {
+	MarshalVT() ([]byte, error)
+	UnmarshalVT([]byte) error
+}
```
marclop, line 25: This interface could be split in 2, so we can accept the smallest possible type in Encode and Decode.

```suggestion
// VTProtoMarshaler ...
type VTProtoMarshaler interface {
	MarshalVT() ([]byte, error)
}

// VTProtoUnmarshaler ...
type VTProtoUnmarshaler interface {
	UnmarshalVT([]byte) error
}
```

codec/vtproto/codec.go
```diff
@@ -0,0 +1,44 @@
...
+
+// VTProtoUnmarshaler expects a struct to have UnmarshalVT() method that
+// parses a given byte slice and places the decoded results in the object.
+type VTProtoUnmarshaler interface {
+	UnmarshalVT([]byte) error
+}
+
+// Codec is a composite of Encoder and Decoder.
+type Codec struct {
+}
```
marclop, line 34:
```suggestion
type Codec struct {}
```

marclop (approved):
LGTM!

kyungeunni:
Reverted changes specifying type instead of `any` for the Encode and Decode as it should implement [Codec interface](https://github.com/elastic/apm-data/blob/e51edfb233ad7ce99a87c9608e52201f98691a30/codec/codec.go#L28-L44)

PR #85, opened 2023-06-23
Add `received` to Event

endorama:
As part of data ingestion one interesting element we want to keep track of is when a certain event has been received by the ingestion backend.

This timestamp is not user-controlled like `@timestamp` and can guarantee us a certain timestamp for event reception, the step before ingestion (which is registered as `event.ingested`).

This PR adds the `event.received` field to protobuf Event and `modeljson.Event`.
This field value will be empty in `modeljson.Event` if it's not set in protobuf `Event`.

Tests have been adjusted to account for the additional field.


axw (approved):
LGTM!

model/modelpb/event.pb.json_test.go
```diff
@@ -64,14 +74,16 @@ func TestEventToModelJSON(t *testing.T) {
...
 			},
 		},
 	}
 	for name, tc := range testCases {
 		t.Run(name, func(t *testing.T) {
 			var out modeljson.Event
 			tc.proto.toModelJSON(&out)
-			diff := cmp.Diff(*tc.expected, out)
+			diff := cmp.Diff(*tc.expected, out,
+				cmpopts.IgnoreUnexported(modeljson.Time{}))
```
kruskall, line 86: **issue**: since we're adding the `received` fields in the above structs, I think we should compare them in the the diff. Could you add a custom comparer ? Something like this should work:
```
cmp.Comparer(func(a modeljson.Time, b modeljson.Time) bool {
	return time.Time(a).Compare(time.Time(b)) == 0
}),
```

kruskall (changes requested):

kruskall (approved):
LGTM! Sorry about the go 1.20 issue :facepalm:

endorama:
@kruskall No worries and thanks for catching the wrong test!

PR #87, opened 2023-06-27
feat: add clonevt methods

kruskall:
Benchmark showed a 50% decrease in cpu time when comparing the two
clone methods:

```
               │  before.txt  │              after.txt              │
               │    sec/op    │   sec/op     vs base                │
Clone/clone-20   15.099µ ± 6%   7.326µ ± 5%  -51.48% (p=0.000 n=10)

               │  before.txt  │              after.txt              │
Clone/clone-20   12.77Ki ± 0%   12.20Ki ± 0%  -4.46% (p=0.000 n=10)

               │ before.txt │             after.txt             │
               │ allocs/op  │ allocs/op   vs base               │
Clone/clone-20   172.0 ± 0%   163.0 ± 0%  -5.23% (p=0.000 n=10)
```

Additionally, since clone already performs a deep copy, we can avoid
cloning the event twice.


axw (approved):
nice!

dmathieu (approved):

PR #88, opened 2023-06-27
feat: add codec benchmarking

kyungeunni:
Benchmark JSON codec and vtproto codec.

Side note: needed to copy over `modelpb/util_test.go` and `fullEvent()` function from one of the test files to generate the full events, couldn't make it sharable as it had a circular dependencies issue with the modelpb <-> testutils packages. Should consider arranging it a bit better later though.

### How to get the benchmark stats:
1. Run a benchmark for encoding or decoding and store it in a text file 
```sh
go test -benchmem -run=^$ -bench ^BenchmarkEncode$ -count=10 github.com/elastic/apm-data/codec > encode.txt
```

2. get stats by using `benchstat`
```sh
benchstat -col /format encode.txt
```

#### Encode
```
goos: darwin
goarch: arm64
pkg: github.com/elastic/apm-data/codec
          │     json     │               vtproto               │
          │    sec/op    │   sec/op     vs base                │
Encode-10   22.861µ ± 2%   2.384µ ± 3%  -89.57% (p=0.000 n=10)

          │     json     │               vtproto                │
          │   bytes/op   │   bytes/op    vs base                │
Encode-10   5.988Ki ± 0%   2.788Ki ± 0%  -53.44% (p=0.000 n=10)

          │     json      │               vtproto                │
          │     B/op      │     B/op      vs base                │
Encode-10   35.848Ki ± 0%   4.949Ki ± 0%  -86.19% (p=0.000 n=10)

          │    json    │              vtproto               │
          │ allocs/op  │ allocs/op   vs base                │
Encode-10   69.00 ± 0%   34.00 ± 0%  -50.72% (p=0.000 n=10)
```

#### Decode
```
goos: darwin
goarch: arm64
pkg: github.com/elastic/apm-data/codec
          │     json     │               vtproto               │
          │    sec/op    │   sec/op     vs base                │
Decode-10   17.401µ ± 2%   6.787µ ± 4%  -61.00% (p=0.000 n=10)

          │     json     │               vtproto               │
Decode-10   13.28Ki ± 0%   14.12Ki ± 0%  +6.35% (p=0.000 n=10)

          │    json    │              vtproto               │
          │ allocs/op  │ allocs/op   vs base                │
Decode-10   323.0 ± 0%   377.0 ± 0%  +16.72% (p=0.000 n=10)

```


kyungeunni:
There seem to be circular dependencies. I will take a look at that and make it ready for review again

codec/json/codec_test.go
```diff
@@ -0,0 +1,51 @@
...
+		var output int64
+		b.RunParallel(func(p *testing.PB) {
+			for p.Next() {
+				by, _ := codec.Encode(*modelpbtest.FullEvent(b))
+				output += int64(len(by))
+			}
+		})
+
+		bytePerOp := float64(output) / float64(b.N)
+		b.ReportMetric(bytePerOp, "bytes/op")
```
marclop, line 39: We seem to be using the same data for encoding, so it'll always be the same size per op. One thing we could do is measure per second, or report the byte per op at the top before the benchmark starts by taking `len(FullEvent)`

codec/json/codec_test.go
```diff
@@ -0,0 +1,55 @@
...
+		// b.ReportMetric(bytePerOp, "bytes/op")
+	})
+
+	b.Run("Decode", func(b *testing.B) {
+		encoded, _ := codec.Encode(testutils.FullEvent(b))
+		var ev *modelpb.APMEvent
+		b.RunParallel(func(p *testing.PB) {
+			for p.Next() {
+				codec.Decode(encoded, ev)
+			}
```
axw, line 52: I was surprised when I saw that the JSON codec was faster in decoding than the Protobuf codec. I think it's because you're not allocating a new event for each iteration in the JSON codec.

I think it may be best if we move both the JSON and Protobuf codecs into `package codec`, and remove `codec/json` and `codec/vtproto`. So we would end up with `codec.JSON` and `codec.VTProto`, or something along those lines.

Then we can use a single benchmark function in this package, and avoid discrepancies like this. I think it's also a bit neater in terms of package organisation. Eventually we should remove the JSON codec anyway.

codec/codec_test.go
```diff
@@ -79,6 +81,63 @@ func TestMetrics(t *testing.T) {
...
+	}{
+		"json": {
+			codec: JSON{},
+		},
+		"vtproto": {
+			codec: VTProto{},
+		},
+	}
+	for name, tc := range testCases {
+		b.Run("format="+name, func(b *testing.B) {
```
marclop, line 97:
```suggestion
		b.Run(name, func(b *testing.B) {
```

marclop, line 97:
```suggestion
		b.Run(name, func(b *testing.B) {
```

codec/codec_test.go
```diff
@@ -79,6 +81,63 @@ func TestMetrics(t *testing.T) {
...
+		"json": {
+			codec: JSON{},
+		},
+		"vtproto": {
+			codec: VTProto{},
+		},
+	}
+	for name, tc := range testCases {
+		encoded, _ := tc.codec.Encode(ev)
+		b.Run("format="+name, func(b *testing.B) {
```
marclop, line 127:
```suggestion
		b.Run(name, func(b *testing.B) {
```

codec/codec_test.go
```diff
@@ -79,6 +81,63 @@ func TestMetrics(t *testing.T) {
 	}, metric, metricdatatest.IgnoreTimestamp())
 }
 
+func BenchmarkEncode(b *testing.B) {
+	ev := fullEvent(b)
+	testCases := map[string]struct {
```
marclop, line 86: This could be simplified to `map[string]Codec`?

codec/codec_test.go
```diff
@@ -79,6 +81,63 @@ func TestMetrics(t *testing.T) {
...
+			})
+			bytePerOp := float64(output.Load()) / float64(b.N)
+			b.ReportMetric(bytePerOp, "bytes/op")
+		})
+	}
+}
+
+func BenchmarkDecode(b *testing.B) {
+	ev := fullEvent(b)
+	testCases := map[string]struct {
```
marclop, line 115: Same here, seems like using a struct with a single field vs the field directly may be more straightforward?

codec/codec_test.go
```diff
@@ -79,6 +81,63 @@ func TestMetrics(t *testing.T) {
...
+		b.Run("format="+name, func(b *testing.B) {
+			b.RunParallel(func(p *testing.PB) {
+				for p.Next() {
+					e := &modelpb.APMEvent{}
+					tc.codec.Decode(encoded, e)
+				}
+			})
+		})
+
+		bytePerOp := float64(len(encoded)) / float64(b.N)
```
marclop, line 136: `encoded` is created at the top and is not incremented for each of the iterations. I think we don't need to divide it by `b.N`.
```suggestion
		bytePerOp := float64(len(encoded))
```

marclop (approved):
LGTM. Just a couple of minor improvements

codec/codec_test.go
kyungeunni: This was to make the stats comparison a bit easier. I can generate one file and compare the two codecs result by using the keys:
```
benchstat -col /format codec.txt
```

The reference is here:  reference is [here](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat#hdr-Configuring_comparisons)

Issue #73, opened 2023-06-19
Update dev docs after models removal and migration to protobuf


kruskall:
Dev docs should be updated to account for protobuf definitions and new `modelpb` package

The old `model` package will be removed

See https://github.com/elastic/apm-data/blob/main/dev_docs/HOW_TO.md

PR #90, opened 2023-06-27
docs: update dev docs for protobuf changes

kruskall:
Update dev docs for protobuf changes

Closes https://github.com/elastic/apm-data/issues/73


dev_docs/HOW_TO.md
```diff
@@ -11,14 +11,17 @@ Adding a new intake field requires changes to both apm-data and apm-server repos
    - RUM v3: [input/elasticapm/internal/modeldecoder/rumv3/model.go](/input/elasticapm/internal/modeldecoder/rumv3/model.go)
 2. Run `make generate` to generate the corresponding `model_generated.go` and JSON specification from the modified `model.go` in step 1.
 3. Run `make update-licenses` to add license header to `model_generated.go` generated in step 2.
-4. Add the new field to the corresponding file in `model/`, e.g. `span.go`.
-   1. Add the field to the struct.
-   2. Modify the `fields` function to include the field.
+4. Add the new field to the corresponding file in `model/proto/`, e.g. `span.proto`.
+   1. Run `make generate-modelpb` to generate the corresponding `span.pb.go`.
+   2. Run `make update-licenses fmt` to add license header to `span.pb.go` generated in step 1.
```
carsonip, line 16: A little bit of ambiguity here, there's also a step 1 in the outer layer
```suggestion
   2. Run `make update-licenses fmt` to add license header to `span.pb.go` generated in the previous step.
```

dev_docs/HOW_TO.md
```diff
@@ -75,9 +78,12 @@ Mapping an OTel field is similar to adding a field to Intake.
 ### 1. Make changes in apm-data (this repo)
 
 1. Modify OTel parsing code in [input/otlp](/input/otlp)
-2. Add the new field to the corresponding file in `model/`, e.g. `span.go`.
-   1. Add the field to the struct.
-   2. Modify the `fields` function to include the field.
+2. Add the new field to the corresponding file in `model/proto/`, e.g. `span.proto`.
+   1. Run `make generate-modelpb` to generate the corresponding `span.pb.go`.
+   2. Run `make update-licenses fmt` to add license header to `span.pb.go` generated in step 1.
```
carsonip, line 83:
```suggestion
   2. Run `make update-licenses fmt` to add license header to `span.pb.go` generated in the previous step.
```

dev_docs/HOW_TO.md
kruskall: Thanks! Good catch :+1:

dev_docs/HOW_TO.md
```diff
@@ -11,14 +11,17 @@ Adding a new intake field requires changes to both apm-data and apm-server repos
    - RUM v3: [input/elasticapm/internal/modeldecoder/rumv3/model.go](/input/elasticapm/internal/modeldecoder/rumv3/model.go)
 2. Run `make generate` to generate the corresponding `model_generated.go` and JSON specification from the modified `model.go` in step 1.
 3. Run `make update-licenses` to add license header to `model_generated.go` generated in step 2.
-4. Add the new field to the corresponding file in `model/`, e.g. `span.go`.
-   1. Add the field to the struct.
-   2. Modify the `fields` function to include the field.
+4. Add the new field to the corresponding file in `model/proto/`, e.g. `span.proto`.
+   1. Run `make generate-modelpb` to generate the corresponding `span.pb.go`.
```
kyungeunni, line 15: Should we also mention that `span_vtproto.pb.go` would also be generated, and we prefer to use the methods from that file when needing to marshal and unmarshal with some brief introduction(as far as I know that was the case)?

kruskall: > Should we also mention that span_vtproto.pb.go would also be generated,

Good catch! :+1: 

>  and we prefer to use the methods from that file when needing to marshal and unmarshal with some brief introduction(as far as I know that was the case)?

I planned to add that as comments/doc to the methods to make it more visible when using the library. Not sure if that approach makes sense, WDYT ?

simitt: IMO this is worth both - adding comments and describing the difference and when to use what in the dev docs.

kyungeunni: I'm with @simitt. As well as the comments on the methods, having an overview or a brief explanation about why we have `{entity}__vtproto.pb.go` files and why we choose to introduce them could help users to utilize them. (Based on my experience as I didn't know about this before Marc told me about it :) )

kruskall: Sorry about that! I've added more sections :bow: 

Let me know what you think!

kyungeunni (approved):
LGTM!

PR #91, opened 2023-06-28
perf: improve batch pool memory usage

kruskall:
Share the pool among processor for more efficient usage. 
Clear the slice before putting it back into the pool.

Because the entire backing array is kept in memory even when slicing, the `modelpb.Batch` is keeping around references to pointers, preventing `APMEvent`s from being garbage collected. This wasn't a big issue before we switched to protobuf models because the `model.Batch` was not a slice of pointers.

### DIFF
--- a/input/elasticapm/processor.go
+++ b/input/elasticapm/processor.go
@@ -42,6 +42,8 @@ var (
 	// ErrQueueFull may be returned by HandleStream when the internal
 	// queue is full.
 	ErrQueueFull = errors.New("queue is full")
+
+	batchPool sync.Pool
 )
 
 const (
@@ -64,7 +66,6 @@ const (
 // the concurrency limit is shared between all the intake endpoints.
 type Processor struct {
 	streamReaderPool sync.Pool
-	batchPool        sync.Pool
 	sem              input.Semaphore
 	logger           *zap.Logger
 	MaxEventSize     int
@@ -334,13 +335,13 @@ func (p *Processor) handleStream(
 		}()
 	}
 	var batch modelpb.Batch
-	if b, ok := p.batchPool.Get().(*modelpb.Batch); ok {
+	if b, ok := batchPool.Get().(*modelpb.Batch); ok {
 		batch = (*b)[:0]
 	}
 	n, readErr = p.readBatch(ctx, baseEvent, batchSize, &batch, sr, result)
 	if n == 0 {
 		// No events to process, return the batch to the pool.
-		p.batchPool.Put(&batch)
+		batchPool.Put(&batch)
 		return readErr
 	}
 	// Async requests are processed in the background and once the batch has
@@ -363,7 +364,12 @@ func (p *Processor) handleStream(
 
 // processBatch processes the batch and returns it to the pool after it's been processed.
 func (p *Processor) processBatch(ctx context.Context, processor modelpb.BatchProcessor, batch *modelpb.Batch) error {
-	defer p.batchPool.Put(batch)
+	defer func() {
+		for i := range *batch {
+			(*batch)[i] = nil
+		}
+		batchPool.Put(batch)
+	}()
 	return processor.ProcessBatch(ctx, batch)
 }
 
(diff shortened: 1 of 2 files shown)


input/elasticapm/processor.go
```diff
@@ -363,7 +364,12 @@ func (p *Processor) handleStream(
 
 // processBatch processes the batch and returns it to the pool after it's been processed.
 func (p *Processor) processBatch(ctx context.Context, processor modelpb.BatchProcessor, batch *modelpb.Batch) error {
-	defer p.batchPool.Put(batch)
+	defer func() {
+		for i := range *batch {
+			(*batch)[i] = nil
+		}
+		batchPool.Put(batch)
+	}()
```
axw, line 372: Should we update apm-data to use vtproto's "pool" feature, and then use ReturnToVTPool/*ProtoFromVTPool functions? ReturnToVTPool will also take care of resetting.

marclop: I just noticed today that this was missing from the current generation. It would be good to add it back.

kruskall: > Should we update apm-data to use vtproto's "pool" feature, and then use ReturnToVTPool/*ProtoFromVTPool functions? ReturnToVTPool will also take care of resetting.

Could you clarify ? AFAICS `modelpb.Batch` is not a protobuf message, just a type alias, so it can't use the protobuf generated methods.

> I just noticed today that this was missing from the current generation. It would be good to add it back.

That's strange, the pool feature is enabled even though not used. Thanks for spotting that!

marclop: The structs need to be specified explicitly. I can add it back later if that's OK with you.

marclop: We could potentially define a Batch ourselves:

```proto
message Batch {
  repeated APMEvent events = 1;
}
```

And let the pool plugin do its thing

kruskall: I can look into it, my comment was mostly due to surprising behaviour in the different usage of the flag.

> We could potentially define a Batch ourselves:
> 
> message Batch {
>   repeated APMEvent events = 1;
> }
> 
> And let the pool plugin do its thing

That would generate a struct but we're not really interested in reusing protobuf models (not in this case at least), we want to reuse the slice to avoid allocating a new one for each batch.

axw: If defining a Batch message type doesn't work out, then how about adding a method to Batch ourselves?

```go
// Reset resets each event in the batch, and truncates the slice to zero length.
func (b *Batch) Reset() {
    for _, event := range *b {
	event.ResetVT()
    }
    *b = (*b)[:0]
}
```

Then

```go
defer batchPool.Put(batch)
defer batch.Reset()
```

axw: I guess that doesn't help with returning the events to the pool though...

kruskall: > ```go
> // Reset resets each event in the batch, and truncates the slice to zero length.
> func (b *Batch) Reset() {
>     for _, event := range *b {
> 	event.ResetVT()
>     }
>     *b = (*b)[:0]
> }
> ```

I might be missing something, why do we want to use a pool of `APMEvent`s ? We're not reusing protobuf models and I don't think that's possible: we rely on the baseevent from APM server which is then cloned before being used by the decoder.

axw: APMEvents will be more expensive to allocate than the slice. Good point about the base events. I think we may need to move away from this approach to enable pooling. We could use a pool and `proto.Merge` the base event on top of an empty event, but reflection will kill us.

Anyway, seems like this is not straightforward at the moment. Let's come back to pooling the events later.

axw, line 372:
```suggestion
	defer batchPool.Put((*batch)[:0])
```

Would this work?

kruskall: I must admin I was under the impression that the continuously allocating the slice would be more expensive.

Agree with what you said, let's come back to this later :+1:

kruskall: No, it wouldn't solve the issue because the underlying array would still be kept fully in memory, keeping references to unused `APMEvent`s

From https://go.dev/blog/slices-intro:

> As mentioned earlier, re-slicing a slice doesn’t make a copy of the underlying array. The full array will be kept in memory until it is no longer referenced.

But since we're reusing the `modelpb.Batch` the unerlying array is always reused and kept alive.

axw (approved):

PR #96, opened 2023-06-30
fix: update handwritten fastjson methods to avoid fallback to reflection

kruskall:
Some handwritten fastjson methods were calling fastjson.Marshal with the wrong types, leading to the library falling back to the slow reflection-based marshaling logic.

### DIFF
--- a/model/internal/modeljson/error.go
+++ b/model/internal/modeljson/error.go
@@ -116,9 +116,16 @@ func (e *Exception) marshalOne(w *fastjson.Writer, offset, parentOffset int) (in
 	if len(e.Stacktrace) != 0 {
 		maybeComma()
 		w.RawString(`"stacktrace":`)
-		if err := fastjson.Marshal(w, e.Stacktrace); err != nil {
-			return -1, err
+		w.RawByte('[')
+		for i, v := range e.Stacktrace {
+			if i != 0 {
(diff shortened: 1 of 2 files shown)


model/internal/modeljson/error.go
```diff
@@ -116,9 +116,16 @@ func (e *Exception) marshalOne(w *fastjson.Writer, offset, parentOffset int) (in
 	if len(e.Stacktrace) != 0 {
 		maybeComma()
 		w.RawString(`"stacktrace":`)
-		if err := fastjson.Marshal(w, e.Stacktrace); err != nil {
-			return -1, err
+		w.RawByte('[')
+		for i, v := range e.Stacktrace {
+			if i != 0 {
+				w.RawByte(',')
+			}
```
marclop, line 123: Is this right? If we're writing a JSON array, the format should be:

```javascript
[
{},
{}
]
```

With the logic above, a comma will be written after the first item has been written, which isn't right:

```javascript
[
{}
{},
]
```

axw: The comma is written before each element, except for before the first element.

marclop: d'uh.... I clearly didn't read this right.

marclop (approved):

PR #98, opened 2023-07-04
Simplify parent and child into strings instead of full-blown messages

dmathieu:
This is part of #47.


model/proto/apmevent.proto
```diff
@@ -76,15 +74,15 @@ message APMEvent {
   Processor processor = 19;
   HTTP http = 20;
   UserAgent user_agent = 21;
-  Parent parent = 22;
+  string parent = 22;
```
marclop, line 77: I think this should be named `parent_id` not just parent?
```suggestion
  string parent_id = 22;
```

dmathieu: Makes sense.
I have also renamed `child` to `childs`.

marclop: shouldn't it be `children`?

dmathieu: It was in the short form, so I didn't extend it. But I can do so.

marclop (approved):

PR #99, opened 2023-07-04
Add make all command

dmathieu:
There are a lot of different make commands in this project, and as I was updating the proto definitions, I ended up having to use many of them (generate-modelpb, update-licenses, fmt for a proto update).
So this `make all` command is a utility tool to make it easier to regenerate and clean up everything.


axw (approved):
Thank you! Maybe later we can add some finer-grained dependencies between tasks, so running "make generate-modelpb" would also add license headers and format the code, without also running fieldalignment and gomodtidy. This is good enough for now though.

PR #103, opened 2023-07-04
Remove unused trace import

dmathieu:
Since https://github.com/elastic/apm-data/pull/100, trace.go isn't used in this file anymore. So the import should be removed.


axw (approved):
LGTM. Perhaps we should use a linter/formatter to do this automatically. Maybe [`buf`](https://buf.build/docs/lint/usage)?

PR #105, opened 2023-07-05
Protobuf linter

dmathieu:
This sets up a linter for protobuf definitions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,12 +13,7 @@ jobs:
       with:
         go-version-file: go.mod
         cache: true
-    - run: make generate
-    - run: make update-licenses
-    - run: make fmt
-    - run: make gomodtidy
-    - run: make fieldalignment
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -18,6 +18,7 @@ jobs:
     - run: make fmt
     - run: make gomodtidy
     - run: make fieldalignment
+    - run: make protolint
```
axw, line 21: Would it make sense to replace all the above with `make all`?

axw (approved):
thank you :)

.github/workflows/ci.yml
dmathieu: That makes sense (note: it also runs `generate-modelpb`, which we didn't run before).

axw: :+1: that sounds desirable, it'll ensure that generated types are up to date.

PR #106, opened 2023-07-05
Return the context in ProcessBatch

dmathieu:
Right now, processors can't communicate with each other, which is something I need to do to start doing tracing within them.
By returning a context after ProcessBatch, the processor can change the context and return a new one, which will be used by every underlying processor.

With that, processors will be able to communicate with each other, and more specifically, one will be able to start a span, and another one to finish it (and possibly every other one, to retrieve the span in context and add attributes).

### DIFF
--- a/input/elasticapm/processor.go
+++ b/input/elasticapm/processor.go
@@ -349,12 +349,14 @@ func (p *Processor) handleStream(
 	if async {
 		go func() {
 			defer p.sem.Release(1)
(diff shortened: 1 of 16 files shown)


axw:
Before we go ahead with this, can you show how it would be used with a code example? Is it possible to do this by wrapping, rather than chaining?

dmathieu:
You are right, using a wrapper processor is a better solution.
Closing in favor of https://github.com/elastic/apm-server/pull/11156

PR #110, opened 2023-07-10
Improve protobuf ip storage

dmathieu:
This improves the protobuf IP storage by replacing the current `string` to a oneoff, with a `fixed32` for ipv4 addresses, and a `bytes` for ipv6.

With this, we can remove some IP addresses parsing.
See https://github.com/elastic/apm-data/pull/42#discussion_r1212572501

```
pkg: github.com/elastic/apm-data/codec
                         │ bench-main  │            bench-branch            │
                         │   sec/op    │   sec/op     vs base               │
Encode/format=json-10      22.14µ ± 5%   21.73µ ± 4%       ~ (p=0.089 n=10)
Encode/format=vtproto-10   2.487µ ± 2%   2.433µ ± 4%       ~ (p=0.138 n=10)
Decode/format=json-10      17.08µ ± 9%   17.75µ ± 9%       ~ (p=0.165 n=10)
Decode/format=vtproto-10   6.879µ ± 7%   6.921µ ± 3%       ~ (p=0.481 n=10)
geomean                    8.967µ        8.978µ       +0.12%

                         │  bench-main  │             bench-branch              │
                         │   bytes/op   │   bytes/op    vs base                 │
Encode/format=json-10      5.988Ki ± 0%   5.988Ki ± 0%       ~ (p=1.000 n=10) ¹
Encode/format=vtproto-10   2.741Ki ± 0%   2.726Ki ± 0%  -0.57% (p=0.000 n=10)
geomean                    4.052Ki        4.040Ki       -0.29%
¹ all samples are equal

pkg: github.com/elastic/apm-data/input/elasticapm/internal/decoder
                   │  bench-main  │         bench-branch          │
                   │    sec/op    │   sec/op     vs base          │
NDStreamDecoder-10   45.74n ± 21%   41.47n ± 5%  ~ (p=0.105 n=10)

pkg: github.com/elastic/apm-data/model/modelpb
                          │ bench-main  │         bench-branch          │
                          │   sec/op    │   sec/op     vs base          │
APMEventToJSON/to-json-10   11.34µ ± 6%   11.37µ ± 1%  ~ (p=0.739 n=10)
```


input/elasticapm/internal/modeldecoder/v2/decoder.go
```diff
@@ -361,26 +360,28 @@ func mapToCloudModel(from contextCloud, cloud *modelpb.Cloud) {
 
 func mapToClientModel(from contextRequest, source **modelpb.Source, client **modelpb.Client) {
 	// http.Request.Headers and http.Request.Socket are only set for backend events.
-	if _, err := netip.ParseAddr((*source).GetIp()); err != nil {
+	if (*source).GetIp() == nil {
 		ip, port := netutil.SplitAddrPort(from.Socket.RemoteAddress.Val)
```
dmathieu, line 364: We could remove one parsing here if we stop relying on netutil, and split the IP/Port from a string directly into an IP.

model/proto/ip.proto
```diff
@@ -0,0 +1,29 @@
...
+syntax = "proto3";
+
+package elastic.apm.v1;
+
+option go_package = "github.com/elastic/apm-data/model/modelpb";
+
+message IP {
+  oneof ip_addr {
+    fixed32 v4 = 1;
+    bytes v6 = 2;
```
kruskall, line 27: **question**: did you measure the storage difference with the previous representation ? 

Also, I wonder if storing the ipv6 in two `fixed64` would have any advantage ?

dmathieu: I've run the following against main:

```
	e := &modelpb.APMEvent{
		Host: &modelpb.Host{
			Ip: []string{"127.0.0.1"},
		},
	}
	c := codec.VTProto{}

	by, err := c.Encode(e)
	if err != nil {
		log.Panic(err)
	}

	fmt.Fprintf(os.Stdout, "%d", len(by))
```

The size is 14 bytes.

And against this branch:

```
	e := &modelpb.APMEvent{
		Host: &modelpb.Host{
			Ip: []*modelpb.IP{
				{
					IpAddr: &modelpb.IP_V4{V4: 2130706433},
				},
			},
		},
	}
	c := codec.VTProto{}

	by, err := c.Encode(e)
	if err != nil {
		log.Panic(err)
	}

	fmt.Fprintf(os.Stdout, "%d", len(by))
```

The size is 10 bytes.

dmathieu: > Also, I wonder if storing the ipv6 in two fixed64 would have any advantage ?

The problem would be that we can't use `repeated` in a `oneoff` message. So we'd have to create a new message type, which may cause more overhead.

axw: OTOH we already have an extra layer due to the `oneof`. Would it be better to remove the `oneof` to avoid reflection? i.e.

```
message IP {
  fixed32 v4 = 1;
  bytes v6 = 2;
}
```

We could always put the low (or high) 4 bytes in V4, and the rest in V6. (Maybe field names would then be low/high (lo/hi)?)

axw (commented):
Generally looks good, just wondering if we should remove the `oneof` to avoid reflection.

model/modelpb/ip.go
```diff
@@ -0,0 +1,92 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package modelpb
+
+import (
+	binary "encoding/binary"
```
axw, line 21:
```suggestion
	"encoding/binary"
```

model/modelpb/ip.go
```diff
@@ -0,0 +1,92 @@
...
+		ip = make([]byte, 4)
+		binary.BigEndian.PutUint32(ip, i.V4)
+	}
+
+	if addr, ok := netip.AddrFromSlice(ip); ok {
+		return addr.String()
+	}
+
+	return ""
+}
```
axw, line 92:
```suggestion
func IP2Addr(i *IP) netip.Addr {
        if addr := i.GetV6(); len(addr) == 16 {
                return netip.AddrFrom16([16]byte(addr))
        }
        if i.GetV4() != 0 {
                var addr [4]byte
                binary.BigEndian.PutUint32(addr[:], i.V4)
                return netip.AddrFrom4(addr)
        }
        return netip.Addr{}
}       
        
func IP2String(i *IP) string {
        addr := IP2Addr(i)
        if addr.IsValid() {
                return addr.String()
        }
        return ""
}       
```

Maybe a little simpler and more efficient? The slice-to-array conversion requires bumping go.mod to 1.20, which we should probably do anyway.

axw (approved):
LGTM, thank you.

model/modelpb/ip.go
dmathieu: See @kruskall's concern in https://github.com/elastic/apm-data/pull/113#pullrequestreview-1528309004

PR #112, opened 2023-07-12
Copy span.id to documents with transaction.id

### LINKED ISSUE #82: Add `span.id` field to transactions
We want to converge span and transaction data models. For that, all transaction documents should also have a `span.id` field (where the field value is a copy of `transaction.id`).
This also will allow better correlation between logs and arbitrary spans (that are either spans or transactions) for OTel use cases.

OTel log records know about the `span.id` they belong to. However, if they are not tied to spans as SpanEvents, there's no way to tell whether the `span.id` in a log record belongs to a span document or a transaction. So, in case of transaction the correlation breaks, because we query for `span.id = XYZ` while transactions do not have a `span.id` field at all.

SylvainJuge:
Fixes #82 

~~Needs https://github.com/elastic/apm-server/pull/11201 in apm-server to be actually used.~~ modifying the intake parts allows to avoid this.


model/modelprocessor/transaction_span_id.go
```diff
@@ -0,0 +1,45 @@
...
+// ProcessBatch copies transaction.id to span.id
+func (SetSpanIdTransaction) ProcessBatch(ctx context.Context, b *modelpb.Batch) error {
+	for i := range *b {
+		setSpanIdTransaction((*b)[i])
+	}
+	return nil
+}
+
+func setSpanIdTransaction(event *modelpb.APMEvent) {
+	if event.Transaction != nil && (event.Span == nil || event.Span.Id == "") {
```
kruskall, line 37:
```suggestion
	if event.Transaction != nil && event.GetSpan().GetId() == "" {
```

kruskall (approved):
LGTM! Just a minor comment

kruskall:
For the lint error: I think `make update-licenses` should fix the issue

input/otlp/traces.go
```diff
@@ -175,6 +175,10 @@ func (c *Consumer) convertSpan(
 			Sampled:             true,
 			RepresentativeCount: representativeCount,
 		}
+		event.Span = &modelpb.Span{
```
kruskall, line 178: **issue**: you need to check that `spanID` is not empty before assigning, otherwise the event output will encode an empty `event.Span` and the tests will fail. (`event.Span` would be no longer nil)

SylvainJuge: this was fixed in d6249c0983c1e8ee75fd7123cba858dedc3fe269

kruskall (approved):

kruskall:
Done! Thanks for the ping!

PR #113, opened 2023-07-13
Require Go 1.20

dmathieu:
This upgrades to Go 1.20, so we can use slice-to-array conversion in #110.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/elastic/apm-data
 
-go 1.19
+go 1.20
 
 require (
 	github.com/gofrs/uuid v4.3.1+incompatible


kruskall (changes requested):
This would require bumping APM Server go version too but we need to support the two latest go version.

axw:
@kruskall that's only relevant for libraries like apm-agent-go, not to apm-server.

axw:
(and we don't have any compatibility guarantees for _this_ library; it exists only to serve apm-server and friends)

kruskall (approved):

kruskall:
Ah sorry about that, I always mix the support policy :face_exhaling:

Issue #48, opened 2023-06-06
Avoid map allocations when mapping modelpb to modeljson 


kruskall:
The new protobuf logic is allocating maps and copying from `structpb.Struct` to `map[string]Any`. 

We don't really need to do this and could investigate passing the `structpb.Struct` type directly which is then encoded to json with a custom marshaling method or something similar. This would improve performance and decrease memory allocations

PR #121, opened 2023-07-18
perf: void map allocations when mapping modelpb to modeljson

kruskall:
Numbers look encouraging:

```
                          │   old.txt   │            new.txt            │
                          │   sec/op    │   sec/op     vs base          │
APMEventToJSON/to-json-20   13.34µ ± 9%   14.27µ ± 4%  ~ (p=0.063 n=10)

                          │   old.txt    │               new.txt               │
APMEventToJSON/to-json-20   29.45Ki ± 0%   26.68Ki ± 0%  -9.39% (p=0.000 n=10)

                          │  old.txt   │              new.txt               │
                          │ allocs/op  │ allocs/op   vs base                │
APMEventToJSON/to-json-20   68.00 ± 0%   52.00 ± 0%  -23.53% (p=0.000 n=10)
```

Closes https://github.com/elastic/apm-data/issues/48


model/modelpb/context.go
```diff
@@ -17,13 +17,12 @@
...
+
 // updateFields transforms in, returning a copy with sanitized keys,
 // suitable for storing as "custom" in transaction and error documents.
-func updateFields(in map[string]any) {
-	for k, v := range in {
-		if lk := sanitizeLabelKey(k); lk != k {
-			delete(in, k)
-			in[lk] = v
-		}
+func updateFields(in []*common.KeyValue) {
```
axw, line 24: Could we do this sanitisation inline in the MarshalFastJSON methods in modeljson?

axw (commented):
Looks good overall.

Re https://github.com/elastic/apm-data/issues/47#issuecomment-1640228125, I think we could force `http.request.env` and probably `http.request.cookies` values to strings now. Let's see what the agent devs say.

The split between modelpb and common feels a little awkward. Would it be feasible to break the dependency from `modelpb -> modeljson`, and add a dependency the other way around? We would need to move the translation to modeljson. Maybe like this:

1. Generate all protobuf types in modelpb
2. Introduce package `model/modeljson`, moving all the `*.json.go` files from modelpb
3. Move `model/internal/modeljson` to `model/modeljson/internal` (or something like that)
4. Remove `modelpb.APMEvent.MarshalFastJSON` and `modelpb.APMEvent.MarshalJSON`, and add a top-level function to `model/modeljson` for marshalling an event:

```go
package modeljson

func MarshalAPMEvent(e *modelpb.APMEvent, w *fastjson.Writer) error
```

WDYT?

kruskall:
> Introduce package model/modeljson, moving all the *.json.go files from modelpb

This doesn't work: `cannot define new methods on non-local type`

axw:
>> Introduce package model/modeljson, moving all the *.json.go files from modelpb
>
> This doesn't work: cannot define new methods on non-local type

I'm not sure I understand what you mean by that. As part of moving the files, we would need to stop adding methods to the types.

e.g. `func (a *Agent) toModelJSON(out *modeljson.Agent)` would become something like `func AgentModelJSON(in *modelpb.Agent, out *modeljson.Agent)`

model/modelpb/context.go
kruskall: Not all of the key values seem to be sanitized so that would change behaviour

kruskall:
Pushed some changes!

axw (commented):
Looks good, thanks! Just a couple of leftovers, and I'm not sure about the comment you left about sanitisation. It looks fine to me on the surface.

Makefile
```diff
@@ -17,7 +17,7 @@ generate:
 	go generate ./...
 
 fieldalignment:
-	go run golang.org/x/tools/go/analysis/passes/fieldalignment/cmd/fieldalignment@v0.4.0 -test=false $(shell go list ./... | grep -v modeldecoder/generator | grep -v test | grep -v model/modelpb)
+	go run golang.org/x/tools/go/analysis/passes/fieldalignment/cmd/fieldalignment@v0.4.0 -test=false $(shell go list ./... | grep -v modeldecoder/generator | grep -v test | grep -v model/modelpb | grep -v model/common)
```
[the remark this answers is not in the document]
kruskall: Yes, good catch!

model/proto/buf.yaml
```diff
@@ -4,3 +4,4 @@ lint:
     - DEFAULT
   except:
     - PACKAGE_DIRECTORY_MATCH
+    - PACKAGE_SAME_GO_PACKAGE
```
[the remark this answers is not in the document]
kruskall: Yep!

model/modelpb/context.go
kruskall: Not all `[]*modelpb.KeyValue`s are sanitized. If we apply the sanitisation inside the marshaling method it would apply to every field.

axw: I see, thanks for explaining. We could address that with a named type, like `type SanitizedKeyValues []*KeyValue`, with a custom `MarshalFastJSON` method. What you have implemented is fine for now though.

Issue #118, opened 2023-07-17
Review protobuf int size and signed/unsigned usage


dmathieu:
I am going to use this issue to verify the size and signed/unsigned for integers in protobuf.
Below is the list of all ints in the proto definitions. For each of them, I will validate the definition with the ingest pipeline and the json decoder and leave comments on the issue.

Note: this only looks at the integers. The floats/double aren't in here.

Once each int is validated and possibly fixed, this issue will be closed.
See https://github.com/elastic/apm-data/issues/47

## client.proto

* [x] **port**

https://github.com/elastic/apm-data/blob/e5765b8f8d8992d4360231ce86d5f57a8d637366/model/proto/client.proto#L29

The maximum port number is 65 535, which is way lower than what an uint32 can carry.
Protobuf doesn't allow setting int16, so this type and every other port using uint32 is valid.

## destination.proto

* [x] **port**

https://github.com/elastic/apm-data/blob/e5765b8f8d8992d4360231ce86d5f57a8d637366/model/proto/destination.proto#L26

See port comment in `client.proto` above.

## event.proto

* [x] **severity**

https://github.com/elastic/apm-data/blob/e5765b8f8d8992d4360231ce86d5f57a8d637366/model/proto/event.proto#L39

Defined as int64 in modeljson, same as we have here. So there doesn't seem to be any reason to downgrade this to a lower size.
https://github.com/elastic/apm-data/blob/9c38d236152715cea5dabcfabee6a4f2b35e849d/model/internal/modeljson/event.go#L30

See https://github.com/elastic/apm-data/pull/123

## experience.proto

* [x] **count**

https://github.com/elastic/apm-data/blob/e5765b8f8d8992d4360231ce86d5f57a8d637366/model/proto/experience.proto#L32

This field is defined as `int` in the JSON decoder.
https://github.com/elastic/apm-data/blob/e5765b8f8d8992d4360231ce86d5f57a8d637366/input/elasticapm/internal/modeldecoder/v2/model.go#L1120

See https://github.com/elastic/apm-data/pull/122

## http.proto

* [x] **transfer_size**

https://github.com/elastic/apm-data/blob/e5765b8f8d8992d4360231ce86d5f57a8d637366/model/proto/http.proto#L47

Defined as int64 in modejson
https://github.com/elastic/apm-data/blob/e5765b8f8d8992d4360231ce86d5f57a8d637366/model/internal/modeljson/http.go#L45

* [x] **encoded_body_size**

[...]

PR #122, opened 2023-07-19
Upgrade modeljson ints to int64 if their proto uses int64

dmathieu:
This is part of https://github.com/elastic/apm-data/issues/118.

Some proto definitions are using int64 when their modeljson has been using int.
I'm doing it this way since the flow is modelpb->modeljson.


input/elasticapm/internal/modeldecoder/nullable/nullable.go
```diff
@@ -193,6 +202,31 @@ func (v *Int) Reset() {
 	v.isSet = false
 }
 
+// Int64 stores an int64 value and the
+// information if the value has been set
+type Int64 struct {
```
dmathieu, line 207: There is a `generator/nullable`, but it seems unrelated to this. And running `make generate` doesn't regenerate this file. So while I don't know what that generator does, I guess it doesn't generate this.

dmathieu:
I'm opening the PR this way after discussing it quickly with @kruskall due to the way the transformation happens.
However, for all these attributes, it seems unlikely the count needs to be an int64. So I'd be fine using int32 in the protobuf definitions too.

input/elasticapm/internal/modeldecoder/nullable/nullable.go
dmathieu: Nevermind, the generator seems to only kick in when `nullable.Int64` is used.

axw (approved):

axw:
> However, for all these attributes, it seems unlikely the count needs to be an int64. So I'd be fine using int32 in the protobuf definitions too.

Maybe should, not sure - but I think keeping them aligned makes sense either way, so we could do that as a separate change.

PR #132, opened 2023-07-31
feat: remove debug log otlp events

kruskall:
When ingesting OTLP events we are logging a message at debug level with the whole event.

On local benchmarks this is using ~50% cpu time and ~8% memory (space), besides making the logs harder to read (the message is quite big).

This was here since the first commit so I'm not sure if I'm missing something but I don't think we should log the whole event.


axw:
> On local benchmarks this is using ~50% cpu time and ~8% memory (space), besides making the logs harder to read (the message is quite big).

This is only the case with debug-level logging enabled, right? Why are benchmarks being run with that configuration?

> This was here since the first commit so I'm not sure if I'm missing something but I don't think we should log the whole event.

Certainly not by default (i.e. at info level), but why not at debug level?

This logging has been useful in the past to diagnose OTLP translation issues.

kruskall:
> This is only the case with debug-level logging enabled, right? Why are benchmarks being run with that configuration?

Yes, that's correct. I'm running the benchmarks against apm-server tilt setup which has debug logging enabled.

> Certainly not by default (i.e. at info level), but why not at debug level?

I guess I'm comparing this to the intake/rum ingestion process. AFAIK we are only doing this for otel and it's surprising behaviour.
I do have a few concerns regarding potential leaks/side effects, for example wouldn't this make any processor redacting PII or securty keys, headers, etc. pointless because the data is logged as soon as it is received by the otlp processor ?

axw:
>> This is only the case with debug-level logging enabled, right? Why are benchmarks being run with that configuration?
>
> Yes, that's correct. I'm running the benchmarks against apm-server tilt setup which has debug logging enabled.

OK, but we also have the option of disabling that, so the points in the PR description are moot.

>> Certainly not by default (i.e. at info level), but why not at debug level?
>
> I guess I'm comparing this to the intake/rum ingestion process. AFAIK we are only doing this for otel and it's surprising behaviour.

Fair point.

> I do have a few concerns regarding potential leaks/side effects, for example wouldn't this make any processor redacting PII or securty keys, headers, etc. pointless because the data is logged as soon as it is received by the otlp processor ?

I wouldn't say it's pointless, because again, this is only for debug-level. This requires action from the user.

Nevertheless, I agree that it would be safer to not give the option, and instead direct users to direct users to OTel SDK logging and/or using opentelemetry-collector in between with a logging exporter. Before removing the code, would you please identify suitable alternatives to point users at? Maybe for now we just add some links somewhere in `dev_docs` - later on we should probably update the APM guide's troubleshooting section.

Issue #129, opened 2023-07-28
Consider using uint64 to store timestamps instead of timestamppb


lahsivjar:
If `APMEvents` is backed by vtproto's pool then `timestamppb` comes out as one of the most allocation heavy object. This was fixed in `apm-aggregation` by using `uint64` to encode timestamps. If `uint64` suits all our needs, we should consider promoting that package to `apm-data`: https://github.com/elastic/apm-aggregation/tree/main/aggregators/internal/timestamppb

kruskall:
Closed by https://github.com/elastic/apm-data/pull/136

PR #134, opened 2023-08-01
feat: migrate proto timestamp to uint64

kruskall:
Replace `timestamppb` with `uint64` in protobuf models

Update modeljson to use `uint64` instead of `time.Time` under the hood.

Update input models to store `int64` to avoid int->time->int conversion.

Add utility methods for usability.

Caveat: we lose the ability to ingest timestamps before the unix epoch or after the year 2262 (sorry, future apm developers).

Closes https://github.com/elastic/apm-data/issues/129


model/proto/event.proto
```diff
@@ -38,5 +37,5 @@ message Event {
   google.protobuf.Duration duration = 8;
   uint64 severity = 9;
 
-  google.protobuf.Timestamp received = 10;
+  uint64 received = 10;
```
axw, line 40: Could you please add a comment to this noting that it's nanoseconds since epoch

model/proto/apmevent.proto
```diff
@@ -52,7 +51,7 @@ import "useragent.proto";
 option go_package = "github.com/elastic/apm-data/model/modelpb";
 
 message APMEvent {
-  google.protobuf.Timestamp timestamp = 1;
+  uint64 timestamp = 1;
```
axw, line 54: Could you please add a comment to this noting that it's nanoseconds since epoch

model/modelpb/timestamp.go
```diff
@@ -0,0 +1,49 @@
...
+
+func PBTimestampAdd(t uint64, d time.Duration) uint64 {
+	ns := uint64(d.Nanoseconds())
+	return t + ns
+}
+
+func PBTimestampSub(t uint64, tt time.Time) time.Duration {
+	b := TimeToPBTimestamp(tt)
+	return time.Duration(t - b)
+}
```
axw, line 49: I don't love rebuilding `time.Time` functions, unless there's a significant performance benefit. Would it be reasonable to just have two functions, to convert to/from `time.Time`, and perform all time math using the usual `time.Time` methods? e.g.

```go
// FromTime converts a time.Time to uint64 nanoseconds since Unix epoch.
func FromTime(t time.Time) uint64

// ToTime converts uint64 nanoseconds since Unix epoch to a time.Time.
func ToTime(v uint64) time.Time
```

model/modelpb/timestamp.go
```diff
@@ -0,0 +1,49 @@
...
+import "time"
+
+// TimeToPBTimestamp encodes a time.Time to Unix epoch nanos in uint64 for protobuf.
+func TimeToPBTimestamp(t time.Time) uint64 {
+	return uint64(t.UnixNano())
+}
+
+// PBTimestampToTime decodes a uint64 of Unix epoch nanos to a time.Time for protobuf.
+func PBTimestampToTime(timestamp uint64) time.Time {
+	return time.Unix(0, int64(timestamp)).UTC()
```
axw, line 29: Should we be passing in seconds & nanoseconds, so we don't lose so much precision?

model/modeljson/internal/document.go
```diff
@@ -68,25 +68,21 @@ type Document struct {
...
+	Timestamp  Time        `json:"@timestamp"`
 }
 
-type Time time.Time
+type Time uint64
 
 func (t Time) MarshalFastJSON(w *fastjson.Writer) error {
 	w.RawByte('"')
-	w.Time(time.Time(t), timestampFormat)
+	w.Time(time.Unix(0, int64(t)).UTC(), timestampFormat)
```
axw, line 81: Should we use PBTimestampToTime here?

Alternatively I think we could leave `modeljson` alone, and convert to time.Time in the caller.

input/otlp/metrics.go
```diff
@@ -284,8 +283,8 @@ func histogramSample(bucketCounts pcommon.UInt64Slice, explicitBounds pcommon.Fl
 type metricsets map[metricsetKey]metricset
 
 type metricsetKey struct {
-	timestamp time.Time
 	signature string // combination of all attributes
+	timestamp uint64
```
axw, line 287: did this need to change?

marclop:
Is this still relevant?

PR #136, opened 2023-08-07
feat: update protobuf models to use uint64 for timestamps

kruskall:
Smaller scope of https://github.com/elastic/apm-data/pull/134

Update protobuf models to store timestamp as uint64.

We can discuss modeljson and input models later, this should be enough to update the aggregation code.


model/modelpb/timestamp.go
```diff
@@ -0,0 +1,30 @@
...
+import "time"
+
+// FromTime converts a time.Time to uint64 nanoseconds since Unix epoch.
+func FromTime(t time.Time) uint64 {
+	return uint64(t.UnixNano())
+}
+
+// ToTime converts uint64 nanoseconds since Unix epoch to a time.Time.
+func ToTime(v uint64) time.Time {
+	return time.Unix(0, int64(v)).UTC()
```
axw, line 29:
```suggestion
	return time.Unix(int64(v/uint64(time.Second)), int64(v%uint64(time.Second))).UTC()
```

`int64(v)` may overflow

axw (approved):
LGTM, just one issue

PR #139, opened 2023-08-10
Add new attribute mappings

stevejgordon:
Adds support for new attributes defined in more recent semantic conversion schema versions. This initial set is needed to support attributes emitted by the .NET Elasticsearch client instrumentation, which conforms to `https://opentelemetry.io/schemas/1.21.0`. The `go.opentelemetry.io/collector/semconv/v1.5.0` package does not include these attributes, and the latest version only includes up to `1.18.0`. For now, I've included the new attributes directly in our mapping code and updated them to ensure both old and new attribute names are supported. In the long term, we need to discuss how we handle other new attributes and have a strategy for mapping new schema versions as they are made stable and released.

### DIFF
--- a/input/otlp/traces.go
+++ b/input/otlp/traces.go
@@ -70,6 +70,12 @@ const (
 	attributeNetworkMNC               = "net.host.carrier.mnc"
 	attributeNetworkCarrierName       = "net.host.carrier.name"
 	attributeNetworkICC               = "net.host.carrier.icc"
+	attributeHttpRequestMethod        = "http.request.method"
+	attributeHttpResponseStatusCode   = "http.response.status_code"
+	attributeServerAddress            = "server.address"
+	attributeServerPort               = "server.port"
+	attributeUrlFull                  = "url.full"
+	attributeUserAgentOriginal        = "user_agent.original"
 )
 
 // ConsumeTraces consumes OpenTelemetry trace data,
@@ -257,11 +263,11 @@ func TranslateTransaction(
 			setLabel(k, event, ifaceAttributeValue(v))
 		case pcommon.ValueTypeInt:
 			switch kDots {
-			case semconv.AttributeHTTPStatusCode:
+			case semconv.AttributeHTTPStatusCode, attributeHttpResponseStatusCode:
 				isHTTP = true
 				httpResponse.StatusCode = uint32(v.Int())
 				http.Response = &httpResponse
-			case semconv.AttributeNetPeerPort:
+			case semconv.AttributeNetPeerPort, attributeServerPort:
 				event.Source = populateNil(event.Source)
 				event.Source.Port = uint32(v.Int())
 			case semconv.AttributeNetHostPort:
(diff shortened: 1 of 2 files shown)


stevejgordon:
Thanks for the approval, @kruskall. I rebased this so that the commit was signed. What's the process for getting this merged?

input/otlp/traces.go
```diff
@@ -257,11 +263,11 @@ func TranslateTransaction(
 			setLabel(k, event, ifaceAttributeValue(v))
 		case pcommon.ValueTypeInt:
 			switch kDots {
-			case semconv.AttributeHTTPStatusCode:
+			case semconv.AttributeHTTPStatusCode, attributeHttpResponseStatusCode:
 				isHTTP = true
 				httpResponse.StatusCode = uint32(v.Int())
 				http.Response = &httpResponse
-			case semconv.AttributeNetPeerPort:
+			case semconv.AttributeNetPeerPort, attributeServerPort:
```
axw, line 270: We assume that transactions represent HTTP server operations, so should follow https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/http/#http-server-semantic-conventions

Note that we are setting the `event.source.port` field from this value. I don't think we want to set `event.source.port = server.port` -- `event.source.port` is the port of the client.

input/otlp/traces.go
```diff
@@ -325,7 +331,7 @@ func TranslateTransaction(
 				if ip, err := modelpb.ParseIP(stringval); err == nil {
 					event.Source.Ip = ip
 				}
-			case semconv.AttributeNetPeerName:
+			case semconv.AttributeNetPeerName, attributeServerAddress:
```
axw, line 334: As above, I don't think it's appropriate to set `event.source.domain = server.address`

axw (changes requested):
@stevejgordon I think the changes for TranslateSpan are fine, but there's a couple of issues in TranslateTransaction. Can you please check those? The `server.*` fields are mentioned in the 1.21 schema here: https://github.com/open-telemetry/semantic-conventions/blob/4bbb8c907402caa90bc077214e8a2c78807c1ab9/schemas/1.21.0#L15-L19

Adding attributes for a newer semver version is fine as long as we keep support for the older version. Longer term we need to figure out something sustainable, but we don't need to solve that today.

input/otlp/traces.go
```diff
@@ -277,11 +283,11 @@ func TranslateTransaction(
 			stringval := truncate(v.Str())
 			switch kDots {
 			// http.*
-			case semconv.AttributeHTTPMethod:
+			case semconv.AttributeHTTPMethod, attributeHttpRequestMethod:
 				isHTTP = true
 				httpRequest.Method = stringval
 				http.Request = &httpRequest
-			case semconv.AttributeHTTPURL, semconv.AttributeHTTPTarget, "http.path":
+			case semconv.AttributeHTTPURL, semconv.AttributeHTTPTarget, "http.path", attributeUrlFull:
```
axw, line 290: Can we move `attributeUrlFull` to its own "// url.*" block?

axw (approved):
@stevejgordon thanks, LGTM - just one minor comment.

input/otlp/traces.go
stevejgordon: Sure

axw (approved):

Issue #131, opened 2023-07-28
Remove lint:ignore in modeldecoder/generator/slice.go


dmathieu:
Update the modeljson generator to remove the redundant `minValue` check from the generated Go code for unsigned integers, while keeping the validation in the generated JSON schema.

So we can remove the nolint in https://github.com/elastic/apm-data/blob/main/input/elasticapm/internal/modeldecoder/generator/slice.go

See https://github.com/elastic/apm-data/pull/123#discussion_r1274969475

PR #123, opened 2023-07-19
Switch ints that can't be negative to uints

dmathieu:
I am opening this PR as part of the #118 work.

```
Running benchstat
goos: darwin
goarch: arm64
pkg: github.com/elastic/apm-data/codec
                         │ bench-main  │            bench-branch            │
                         │   sec/op    │   sec/op     vs base               │
Encode/format=json-10      22.04µ ± 2%   22.19µ ± 3%       ~ (p=0.271 n=10)
Encode/format=vtproto-10   1.357µ ± 6%   1.359µ ± 2%       ~ (p=0.971 n=10)
Decode/format=json-10      16.23µ ± 7%   16.28µ ± 9%       ~ (p=0.684 n=10)
Decode/format=vtproto-10   6.423µ ± 3%   6.430µ ± 3%       ~ (p=0.912 n=10)
geomean                    7.472µ        7.496µ       +0.31%

                         │  bench-main  │             bench-branch              │
                         │   bytes/op   │   bytes/op    vs base                 │
Encode/format=json-10      6.169Ki ± 0%   6.169Ki ± 0%       ~ (p=1.000 n=10) ¹
Encode/format=vtproto-10   2.702Ki ± 0%   2.702Ki ± 0%       ~ (p=1.000 n=10) ¹
geomean                    4.083Ki        4.083Ki       +0.00%
¹ all samples are equal

pkg: github.com/elastic/apm-data/input/elasticapm/internal/decoder
                   │ bench-main  │         bench-branch          │
                   │   sec/op    │   sec/op     vs base          │
NDStreamDecoder-10   42.51n ± 7%   43.08n ± 7%  ~ (p=0.529 n=10)

pkg: github.com/elastic/apm-data/model/modelpb
                          │ bench-main  │         bench-branch          │
                          │   sec/op    │   sec/op     vs base          │
APMEventToJSON/to-json-10   11.62µ ± 3%   11.66µ ± 6%  ~ (p=0.684 n=10)
```

PR #145, opened 2023-08-25
refactor: remove minvals check for uints

kruskall:
Remove minvals check for uints

Keep min=0 in the schema and remove lint:ignore rule from the generator

Closes https://github.com/elastic/apm-data/issues/131

### DIFF
--- a/input/elasticapm/internal/modeldecoder/generator/slice.go
+++ b/input/elasticapm/internal/modeldecoder/generator/slice.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"go/types"
 	"io"
-	"strings"
 
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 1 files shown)


dmathieu:
The original idea for this fix was to infer a minimum value for the generated JSON schema.
https://github.com/elastic/apm-data/pull/123#discussion_r1274969475

input/elasticapm/internal/modeldecoder/generator/slice.go
```diff
@@ -142,6 +142,9 @@ func generateJSONPropertySlice(info *fieldInfo, parent *property, child *propert
 	default:
 		return fmt.Errorf("unhandled slice item type %s", itemsType)
 	}
+	if items.Min == "" && strings.HasPrefix(itemType.String(), "uint") {
```
axw, line 145: Perhaps check if the underlying type is a `*types.Basic`, and use https://pkg.go.dev/go/types#Basic.Info to check if it's unsigned? Something like:

```go
if basic, ok := itemType.Underlying().(*types.Basic); ok {
	if items.Min == "" && (basic.Info() & types.IsUnsigned) != 0 {
		items.Min = json.Number("0")
	}
}
```

Not a big deal though. Feels slightly cleaner to me, but it's a few more lines.

axw (approved):
thank you!

dmathieu:
With the check for `types.IsUnsigned`, shouldn't we bring back `minVal=0` for jsonschema validation? (that's what I meant in my comment earlier).

axw:
@dmathieu are you asking why there's no diff on the generated JSON Schema? That's because it already had the minimum value property, and still does: https://github.com/elastic/apm-data/blob/3d5c2a73763d32b36efcbb048c666ce208fd2df5/input/elasticapm/docs/spec/v2/metricset.json#L91

Issue #130, opened 2023-07-28
Use `modelpb.<Type>FromVTPool` wherever possible


marclop:
## Description

Since we added back pooling in #128, we should start using the pooled `modelpb.<Type>FromVTPool` wherever possible and indicate that clients should use `ReturnToVTPool` after they're done processing an event.

marclop:
I started looking at changing at least the Intake/v2 decoding.

marclop:
I started looking at this but stopped since I don't have the bandwidth to take this on ATM. Feel free to pick this up as desired.

axw:
Is there anything left to do here?

PR #155, opened 2023-09-20
feat: improve vtpool and cloning usage

kruskall:
Part 1 of https://github.com/elastic/apm-data/pull/152

Related to https://github.com/elastic/apm-data/issues/130

Migrate to vtprotobuf fork, regenerate modelpb and return batch events to the vtpool.

The additional `go.mod` is needed because the fork is still using the original package name so we need a replace directive.


dmathieu:
Should the renovate config be updated to handle `tools/go.mod`?

dmathieu:
> That go.mod is a temporary solution until the upstream PRs are merged in and we can get rid of the fork.

🤞

marclop (approved):
LGTM

PR #158, opened 2023-09-21
feat: migrate decoding and processors to vtpool

kruskall:
Part 2 of https://github.com/elastic/apm-data/pull/152

Migrate model processors and decoder to vtpool 

Blocked by https://github.com/elastic/apm-data/pull/155

Related to https://github.com/elastic/apm-data/issues/130

PR #160, opened 2023-09-22
feat: improve slice usage from pooled objects

kruskall:
Followup to https://github.com/elastic/apm-data/pull/158

Part 3 of https://github.com/elastic/apm-data/pull/152

Objects retrieved from the vtpool have preallocated empty slices. We should reuse them instead of creating new one.

Related to https://github.com/elastic/apm-data/issues/130

### DIFF
--- a/input/elasticapm/internal/modeldecoder/modeldecoderutil/slice.go
+++ b/input/elasticapm/internal/modeldecoder/modeldecoderutil/slice.go
@@ -17,8 +17,11 @@
 
 package modeldecoderutil
 
-func Reslice[Slice ~[]model, model any](slice Slice, want int, newFn func() model) Slice {
-	if diff := want - cap(slice); diff > 0 {
+// Reslice increases the slice's capacity, if necessary, to match n.
+// If specified, the newFn function is used to create the elements
+// to populate the additional space appended to the slice.
+func Reslice[Slice ~[]model, model any](slice Slice, n int, newFn func() model) Slice {
+	if diff := n - cap(slice); diff > 0 {
 		extra := make([]model, diff)
 		if newFn != nil {
 			for i := range extra {
@@ -27,5 +30,5 @@ func Reslice[Slice ~[]model, model any](slice Slice, want int, newFn func() mode
 		}
 		slice = append([]model(slice)[:cap(slice)], extra...)
 	}
-	return slice[:want]
+	return slice[:n]
 }


input/elasticapm/internal/modeldecoder/modeldecoderutil/http.go
```diff
@@ -45,18 +45,18 @@ func HTTPHeadersToStructPb(h http.Header) *structpb.Struct {
 	return nil
 }
 
-func HTTPHeadersToModelpb(h http.Header) []*modelpb.HTTPHeader {
+func HTTPHeadersToModelpb(h http.Header, out []*modelpb.HTTPHeader) []*modelpb.HTTPHeader {
```
marclop, line 48: This doesn't seem right, we're passing out and also returning out?

input/elasticapm/internal/modeldecoder/modeldecoderutil/keyvaluepb.go
```diff
@@ -22,20 +22,21 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
-func ToKv(m map[string]any) []*modelpb.KeyValue {
-	nm := normalizeMap(m)
-	if len(nm) == 0 {
+func ToKv(m map[string]any, out []*modelpb.KeyValue) []*modelpb.KeyValue {
```
marclop, line 25: Same here.

marclop (commented):
were you able to run any benchmarks with this change?

input/elasticapm/internal/modeldecoder/modeldecoderutil/http.go
marclop: I find this API a bit confusing, I think it'd be best to do one or the other. If we want to accept the `out` as an argument, then we can pass `out *[]*modelpb.HTTPHeader`

carsonip: I guess this signature looks similar to `s = append(s, struct{})`, so that's an argument for this pattern. But we'll have to remember to use the return value.

carsonip: But for conversion purposes like this specific one, I think @marclop 's comment makes sense. It will be clearer.

marclop: Thanks for chiming in. I was unaware of the widespread apdoption of this pattern in the current Go stdlib. I still think it's clearer and simpler to accept a single argument. If we leave it as is, I would prefer if we documented the behaviour since there are currently no godocs for these functions.

carsonip (approved):

PR #165, opened 2023-09-28
refactor: implement reslice with stdlib slices package 

kruskall:
Simplify logic using `slices.Grow` func.

Avoid using vtpool for slice elements since they are never returned to the pool but they are just reset.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,7 +23,7 @@ require (
 	go.uber.org/zap v1.27.0
 	golang.org/x/sync v0.7.0
 	golang.org/x/tools v0.20.0
-	google.golang.org/grpc v1.63.0
+	google.golang.org/grpc v1.63.2
 	google.golang.org/protobuf v1.33.0
 )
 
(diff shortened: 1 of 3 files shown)


input/elasticapm/internal/modeldecoder/modeldecoderutil/slice.go
```diff
@@ -31,6 +32,7 @@ func Reslice[Slice ~[]model, model any](slice Slice, n int, newFn func() model)
 		slice = append([]model(slice)[:cap(slice)], make([]model, diff)...)
 		if newFn != nil {
 			// extend the slice to its capacity
```
carsonip, line 34:
```suggestion
			// extend the slice to its capacity only to initialize the new elements, and the length needs to be reset before returning
```

input/elasticapm/internal/modeldecoder/modeldecoderutil/slice.go
```diff
@@ -17,10 +17,11 @@
 
 package modeldecoderutil
 
-// Reslice increases the slice's capacity, if necessary, to match n.
-// If specified, the newFn function is used to create the elements
-// to populate the additional space appended to the slice.
+// Reslice increases the slice's capacity, if necessary, to guarantee space for n elements.
```
carsonip, line 20:
```suggestion
// Reslice increases the slice's capacity and length, if necessary, to guarantee space for a total of n elements.
```

input/elasticapm/internal/modeldecoder/modeldecoderutil/slice.go
```diff
@@ -31,6 +32,7 @@ func Reslice[Slice ~[]model, model any](slice Slice, n int, newFn func() model)
 		slice = append([]model(slice)[:cap(slice)], make([]model, diff)...)
 		if newFn != nil {
 			// extend the slice to its capacity
+			// Note: reset the slice length to n before returning
```
carsonip, line 35:
```suggestion
```

input/elasticapm/internal/modeldecoder/modeldecoderutil/slice.go
```diff
@@ -40,5 +42,9 @@ func Reslice[Slice ~[]model, model any](slice Slice, n int, newFn func() model)
 		}
 	}
 
+	// slice has enough capacity to hold n elements and there is no need to grow the slice.
```
carsonip, line 45:
```suggestion
```
This line is misleading as the return also handles the case of growing the slice.

PR #167, opened 2023-10-12
refactor: drop gofrs/uuid dependency

kruskall:
Not worth it to add a library for a single method call


input/otlp/exceptions.go
```diff
@@ -215,3 +215,29 @@ func parseJavaStacktraceFrame(s string, out *modelpb.Exception) error {
 func isNotTab(r rune) bool {
 	return r != '\t'
 }
+
+func newUUID() (string, error) {
```
axw, line 219: It's not important that it has a UUID format, just that it's a unique ID. Maybe simplify the code below to just hex-encode the random bytes, rather than formatting as UUIDv4.

axw (approved):
LGTM, just one thought about simplifying

carsonip (approved):

PR #170, opened 2023-10-16
benchdiff

v1v:
Test from a non forked PR

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       checks: write
-      contents: write
+      pull-requests: write
     steps:
     - uses: actions/checkout@v4
       with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,9 @@ jobs:
 
   run-benchdiff:
     runs-on: ubuntu-latest
-    permissions: write-all
+    permissions:
+      # for the GitHub comment with the outcome
+      pull-requests: write
     steps:
     - uses: actions/checkout@v4
       with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,26 +36,11 @@ jobs:
 
   run-benchdiff:
     runs-on: ubuntu-latest
-    permissions:
-      # for the GitHub comment with the outcome
-      pull-requests: write
+    permissions: write-all
     steps:
     - uses: actions/checkout@v4
       with:
         fetch-depth: 0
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -36,3 +36,34 @@ jobs:
...
+    - uses: WillAbides/benchdiff-action@4d1d267fa96763646dd7c0d58e242817ce392c61
+      id: benchdiff
+      with:
+        benchdiff_version: 0.9.1
+        status_sha: ${{ github.sha }}
+        status_name: benchdiff-result
+        status_on_degraded: neutral
+        # See https://github.com/WillAbides/benchdiff
+        benchdiff_args: |
+          --base-ref=origin/main
```
v1v, line 61: I've decided to use always `main` to compare with, if needed we could make this dynamic using `github.base_ref` if running on a Pull Request, otherwise `origin/main`.

.github/workflows/ci.yml
```diff
@@ -36,3 +36,34 @@ jobs:
...
+          --cpu=1,2
+          --count=5
+          --warmup-count=1
+          --warmup-time=10ms
+          --benchtime=100ms
+          --tolerance=20
```
v1v, line 67: >   --tolerance=20     The minimum percent change before a result is considered degraded.

In this particular case if 20% variation then it will report a degradation.

For further details please refers to https://github.com/WillAbides/benchdiff#benchdiff---help-output

.github/workflows/ci.yml
```diff
@@ -36,3 +36,34 @@ jobs:
...
       run: make test
+
+  run-benchdiff:
+    runs-on: ubuntu-latest
+    permissions:
+      checks: write
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
```
v1v, line 47: This is required so it can compare agains a particular base-ref, in this particular case `origin/main`

.github/workflows/ci.yml
```diff
@@ -36,3 +36,34 @@ jobs:
...
+      with:
+        go-version-file: go.mod
+    # Version: https://github.com/WillAbides/benchdiff-action/releases/tag/v0.3.3
+    - uses: WillAbides/benchdiff-action@4d1d267fa96763646dd7c0d58e242817ce392c61
+      id: benchdiff
+      with:
+        benchdiff_version: 0.9.1
+        status_sha: ${{ github.sha }}
+        status_name: benchdiff-result
+        status_on_degraded: neutral
```
v1v, line 58: This could help with reporting a failure if a higher degradation based on the `--tolerance=20` value

reakaleek (approved):

PR #185, opened 2023-11-24
Added support for receiving inferred span children as OTLP span links.

### LINKED ISSUE #175: Map child-ids in OTel attributes
To support inferred spans in OTel-based agents, we need to map child-ids field from OTel attributes to the `child.id` field.

TBD: the OTel attribute name to map from.

JonasKunz:
Closes #175. Span-Links received via OTLP with the attribute `elastic.is_child` set to `true` will be stored as `child.id`s instead of links.

This is currently in draft until naming discussions about the attribute are settled.

I've manually tested this PR with our inferred-spans Java Otel extension:
image

### DIFF
--- a/input/otlp/traces.go
+++ b/input/otlp/traces.go
@@ -1147,13 +1147,19 @@ func translateSpanLinks(out *modelpb.APMEvent, in ptrace.SpanLinkSlice) {
 	if out.Span == nil {
 		out.Span = modelpb.SpanFromVTPool()
 	}
-	out.Span.Links = make([]*modelpb.SpanLink, n)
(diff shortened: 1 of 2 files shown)


JonasKunz:
We've discussed this in the APM agents weekly and decided to stay with the proposed `elastic.is_child` attribute name. I'll take care of also adding this to the agent specs.

simitt:
@jackshirazi ran a program to test this and sent data to an APM Server BC5 in ESS; seeing the waterfall for the trace and the span:
Screenshot 2024-03-20 at 17 32 30

PR #188, opened 2023-12-04
feat: map OTEL's `span.status: Unset` to `event.outcome: success`

kyungeunni:
apm-data translates `Span.status: Unset` as `event.outcome: unknown`. For OTEL, Unset status implies has not been failed/errored. We'd like to map Unset to success so we can calculate the correct [transaction failed rate](https://github.com/elastic/apm/blob/7ef63617e84081139e6b79b5006c035057dc42fb/specs/agents/tracing-transactions.md?plain=1#L40).

The [OTel specification](https://opentelemetry.io/docs/concepts/signals/traces/#span-status) writes the following on the SpanStatus being Unset:

> When an exception is handled, a Span status can be set to Error. Otherwise, a Span status is in the Unset state. By setting a Span status to Unset, the back-end that processes spans can now assign a final status.

So it should be ok for us map the Unset to success when we process the data.

### DIFF
--- a/input/elasticapm/internal/modeldecoder/v2/span_test.go
+++ b/input/elasticapm/internal/modeldecoder/v2/span_test.go
@@ -172,17 +172,18 @@ func TestDecodeMapToSpanModel(t *testing.T) {
 		input.Context.HTTP.StatusCode.Set(http.StatusBadRequest)
 		mapToSpanModel(&input, &out)
 		assert.Equal(t, "failure", out.Event.Outcome)
-		// derive from other fields - success
-		input.Outcome.Reset()
-		input.Context.HTTP.StatusCode.Reset()
-		mapToSpanModel(&input, &out)
-		assert.Equal(t, "success", out.Event.Outcome)
-		// derive from other fields - unknown when not otel
+		// derive from other fields - unknown
 		input.Outcome.Reset()
 		input.OTel.Reset()
 		input.Context.HTTP.StatusCode.Reset()
 		mapToSpanModel(&input, &out)
 		assert.Equal(t, "unknown", out.Event.Outcome)
+		// outcome is success when not assigned and it's otel
+		input.Outcome.Reset()
+		input.OTel.SpanKind.Set(spanKindInternal)
+		input.Context.HTTP.StatusCode.Reset()
+		mapToSpanModel(&input, &out)
+		assert.Equal(t, "success", out.Event.Outcome)
 	})
 
 	t.Run("timestamp", func(t *testing.T) {
--- a/input/elasticapm/internal/modeldecoder/v2/span_test.go
+++ b/input/elasticapm/internal/modeldecoder/v2/span_test.go
@@ -172,8 +172,14 @@ func TestDecodeMapToSpanModel(t *testing.T) {
 		input.Context.HTTP.StatusCode.Set(http.StatusBadRequest)
 		mapToSpanModel(&input, &out)
 		assert.Equal(t, "failure", out.Event.Outcome)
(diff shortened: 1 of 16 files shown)


input/otlp/traces.go
```diff
@@ -468,11 +468,12 @@ func TranslateTransaction(
 		}
 
 		// Set outcome nad result from status code.
+		// However, if outcome was explicitly set to failure, preserve it
 		if statusCode := httpResponse.StatusCode; statusCode > 0 {
-			if event.Event.Outcome == outcomeUnknown {
+			if event.Event.Outcome != outcomeFailure {
```
lahsivjar, line 473: Not sure about this one, should we be overriding an outcome which is explicitly set to some value?

input/otlp/traces.go
```diff
@@ -468,11 +468,12 @@ func TranslateTransaction(
...
 				event.Event.Outcome = serverHTTPStatusCodeOutcome(int(statusCode))
 			}
-			if event.Transaction.Result == "" {
+			if event.Transaction.Result != "Error" {
```
lahsivjar, line 476: Same concern as the previous one, should we be overriding an explicitly set value?

input/otlp/traces.go
```diff
@@ -815,7 +816,7 @@ func TranslateSpan(spanKind ptrace.SpanKind, attributes pcommon.Map, event *mode
 	}
 
 	if isHTTP {
-		if httpResponse.StatusCode > 0 && event.Event.Outcome == outcomeUnknown {
+		if httpResponse.StatusCode > 0 {
```
lahsivjar, line 819: same as others.

lahsivjar (commented):
Not sure if I am missing something here, should we be overriding explicitly set `event.outcome` or `transaction.result`?

input/otlp/traces.go
kyungeunni: Thanks @lahsivjar, My original approach was to set `Success` as default when status was Unset, then translateSpan or translateTransaction to override the value, which kind of treated `Success` as not explicitly set value. However, after reading your comment I realized it might not be the best approach. I've reverted it to check for the Outcome at the end and set it as a success if it's still not assigned.

input/elasticapm/internal/modeldecoder/v2/span_test.go
```diff
@@ -174,9 +174,16 @@ func TestDecodeMapToSpanModel(t *testing.T) {
 		assert.Equal(t, "failure", out.Event.Outcome)
 		// derive from other fields - unknown
 		input.Outcome.Reset()
+		input.OTel.Reset()
```
kyungeunni, line 177: The change was only made to OTel input for now, so I needed to reset the Otel attributes here so we don't get confused why it affects also `elasticapm` input

input/elasticapm/internal/modeldecoder/v2/span_test.go
```diff
@@ -174,9 +174,16 @@ func TestDecodeMapToSpanModel(t *testing.T) {
...
 		input.Context.HTTP.StatusCode.Reset()
 		mapToSpanModel(&input, &out)
 		assert.Equal(t, "unknown", out.Event.Outcome)
+		// outcome is success when not assigned and it's otel
+		input.Outcome.Reset()
+		input.OTel.SpanKind.Set(spanKindInternal)
```
kyungeunni, line 183: The change was only made to OTel input for now, if otel attributes are set, it will use the same logic for mapping the Unset status to success outcome

marclop (approved):

PR #189, opened 2023-12-06
Remove semaphore acquire from span

endorama:
To compute more precise timing on the execution of HandleStream and allow monitoring semaphore operation using other instrumentation, move the span start in `HandleStream` after semaphore acquisition.

This prevents the context to be rewritten when passed to `sem.Acquire`.


endorama:
Closing this PR in favor of #190

Issue #193, opened 2023-12-18
remove async processing


kruskall:
non-blocking ingest behaviour was added because of https://github.com/elastic/apm-server/issues/7872 to be used in the lambda extension.

It was never implemented and apm-server has improved since then. Recently there was some discussion around sync/async and it was proposed to remove it since we are not using it and it's just additional complexity.

cc @axw for feedback since you opened the initial issue

axw:
Given that we never actually made use of it, and the general thrust behind treating OpenTelemetry as first class citizen (with no way to have those clients request such behaviour), I'm good with removing it.

PR #199, opened 2024-01-11
remove async arg

SylvainJuge:
Follow-up of #193 , the `async` parameter was removed, but the `HandleStream` interface was not updated to reflect this.

This currently prevents updating apm-data version used in apm-server.

Once this is merged https://github.com/elastic/apm-server/pull/12386 will be unblocked and must have its version of apm-data updated.

### DIFF
--- a/input/elasticapm/processor.go
+++ b/input/elasticapm/processor.go
@@ -86,7 +86,6 @@ type Config struct {
 type StreamHandler interface {
 	HandleStream(
 		ctx context.Context,
-		async bool,
 		baseEvent *modelpb.APMEvent,
 		stream io.Reader,
 		batchSize int,


simitt (approved):
Apologies for this oversight, and thanks for contributing the fix.

SylvainJuge:
Thanks @simitt for the review, can we merge this now ?

As a follow-up we should probably set-up a CI job that ensures apm-server `main` branch is tested with the current state of apm-data `main` branch, or 100x better an automatic PR is created to update apm-data in apm-server so we have feedback when changes merged in apm-data break something. Given we expect contributions from outside of the apm-server team this is quite likely to happen (my future self is the one that is likely to make this kind of oversight).

simitt:
>or 100x better an automatic PR is created to update apm-data in apm-server so we have feedback when changes merged in apm-data break something

agreed; created https://github.com/elastic/apm-server/issues/12268 recently

PR #202, opened 2024-01-24
Added support for elastic.profiler_stack_trace_ids OTel attribute

### LINKED ISSUE #180: Map and store profiling stacktrace IDs on transactions
https://github.com/elastic/apm-agent-java/issues/3409 will add the capability for transactions to store the IDs of stacktraces captured by the profiler during the corresponding. We will need to map and store these correctly in the APM server.

JonasKunz:
Closes #180 .


kyungeunni (dismissed):
LGTM

input/otlp/traces.go
```diff
@@ -270,7 +270,19 @@ func TranslateTransaction(
 		k := replaceDots(kDots)
 		switch v.Type() {
 		case pcommon.ValueTypeSlice:
-			setLabel(k, event, ifaceAttributeValue(v))
+			switch kDots {
+			case "elastic.profiler_stack_trace_ids":
+				var vSlice = v.Slice()
+				event.Transaction.ProfilerStackTraceIds = make([]string, 0, vSlice.Len())
+				for i := 0; i < vSlice.Len(); i++ {
```
simitt, line 277: use [`range`](https://gobyexample.com/range) instead?

simitt (approved):
left a minor comment, otherwise LGTM

input/otlp/traces.go
carsonip: imo this is specific to OTel and we can't range over it.

input/otlp/traces.go
```diff
@@ -270,7 +270,19 @@ func TranslateTransaction(
 		k := replaceDots(kDots)
 		switch v.Type() {
 		case pcommon.ValueTypeSlice:
-			setLabel(k, event, ifaceAttributeValue(v))
+			switch kDots {
+			case "elastic.profiler_stack_trace_ids":
+				var vSlice = v.Slice()
+				event.Transaction.ProfilerStackTraceIds = make([]string, 0, vSlice.Len())
```
carsonip, line 276: We can potentially save an alloc here by:
1. Using https://pkg.go.dev/slices#Grow (requires go 1.21) or https://pkg.go.dev/golang.org/x/exp/slices (requires a new dep). Since apm-server is on 1.21 I think we should also bump apm-data to 1.21
2. Having a nil check and `make` if nil. Otherwise, let it grow on append, if cap is insufficient.

If (1) sounds good, I can bump go version in a separate PR.

carsonip: https://github.com/elastic/apm-data/pull/215

input/elasticapm/internal/modeldecoder/v2/transaction_test.go
```diff
@@ -636,6 +638,21 @@ func TestDecodeMapToTransactionModel(t *testing.T) {
 			mapToTransactionModel(&input, &event)
 			assert.Equal(t, "CLIENT", event.Span.Kind)
 		})
+
+		t.Run("elastic-profiling-ids", func(t *testing.T) {
+			var input transaction
+			var event modelpb.APMEvent
+			modeldecodertest.SetStructValues(&input, modeldecodertest.DefaultValues())
+			input.Type.Reset()
```
carsonip, line 646:
```suggestion
```

input/elasticapm/internal/modeldecoder/v2/transaction_test.go
```diff
@@ -636,6 +638,21 @@ func TestDecodeMapToTransactionModel(t *testing.T) {
...
+			attrs := map[string]interface{}{
+				"elastic.profiler_stack_trace_ids": []interface{}{"id1", "id2"},
+			}
+			input.OTel.Attributes = attrs
+			input.OTel.SpanKind.Reset()
```
carsonip, line 651:
```suggestion
```

input/otlp/traces.go
carsonip: merge main, import slices, and use `slices.Grow`
```suggestion
				event.Transaction.ProfilerStackTraceIds = slices.Grow(event.Transaction.ProfilerStackTraceIds, vSlice.Len())
```

carsonip (approved):

Issue #177, opened 2023-10-27
Map OTel `code.stacktrace` attribute top-level


AlexanderWert:
Once https://github.com/open-telemetry/semantic-conventions/pull/435 lands, we need to map the OTel `code.stacktrace` attribute top-level (instead of into labels) so we can display stacktraces on spans for OTel-based agents.

SylvainJuge:
Implementation is split in two PRs:
- https://github.com/elastic/apm-data/pull/184 for `apm-data`
- https://github.com/elastic/apm-server/pull/12096 for `apm-server`

PR #203, opened 2024-01-24
Added mapping of experimental Otel telemetry.distro.name and version attributes

### LINKED ISSUE #200: Map new telemetry distro attributes to agent fields for OTel distros
The semantic conventions define two experimental attributes on a span.

- `telemetry.distro.name`
- `telemetry.distro.version`

These fields should be used to define the `agent.Name` and `agent.Version` fields.

The `agent.Name` should be, for example:

`opentelemetry/java/elastic`

and the `agent.Version` is the distro version.

JonasKunz:
Closes #200 .

### DIFF
--- a/input/otlp/metadata.go
+++ b/input/otlp/metadata.go
@@ -289,6 +289,9 @@ func translateResourceMetadata(resource pcommon.Resource, out *modelpb.APMEvent)
 		// devices. APM server should drop this field.
 		case "telemetry.sdk.elastic_export_timestamp":
 			// Do nothing.
+		case "telemetry.distro.name":
+		case "telemetry.distro.version":
+			//distro version & name are handled below and should not end up as labels
 
 		default:
 			if out.Labels == nil {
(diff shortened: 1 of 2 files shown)


input/otlp/metadata.go
```diff
@@ -374,12 +377,31 @@ func translateResourceMetadata(resource pcommon.Resource, out *modelpb.APMEvent)
...
+	if distroNameSet && distroName.Str() != "" {
+		agentLang := "unknown"
+		if out.GetService().GetLanguage().GetName() != "" {
+			agentLang = out.GetService().GetLanguage().GetName()
+		}
+
+		out.Agent.Name = fmt.Sprintf("%s/%s/%s", out.Agent.Name, agentLang, distroName.Str())
+
+		//we intentionally do not want to fallback to the Otel SDK version if we have a distro name, this would only cause confusion
+		out.Agent.Version = "unknown"
```
JonasKunz, line 393: @AlexanderWert @estolfo do you agree that it's better to use `unknown` as `agent.version` instead of the sdk-version if a distro-name was found but no distro-version?

I'm thinking that the ambiguity that in that case `agent.version` could be both the distro version or the sdk-version would cause a lot of confusion if both follow similar versioning schemes

SylvainJuge:
Once we will have "otel native" storage, those fields would be stored as top-level elements.

We will need to modify the UI to make use of those (for example to show a proper Elastic + Otel label on services where it's used), so if we can also store them as top-level elements that would provide a future-proof mapping in the UI.

This is the strategy that was used to add `code.stacktrace` in https://github.com/elastic/apm-data/issues/177

input/otlp/metadata.go
JonasKunz: Agreed on in the agents-weekly meeting.

input/otlp/metadata.go
```diff
@@ -386,12 +389,31 @@ func translateResourceMetadata(resource pcommon.Resource, out *modelpb.APMEvent)
 		// agent.version is a required field.
 		out.Agent.Version = "unknown"
 	}
-	if out.GetService().GetLanguage().GetName() != "" {
-		if out.Agent == nil {
-			out.Agent = modelpb.AgentFromVTPool()
```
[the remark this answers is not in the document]
JonasKunz: I would have assumed that `out.GetAgent().getName()` always returns `""` when `out.Agent` is nil.
Therefore, this branch is guaranteed to initialize the `out.Agent`.

I'll move the initialization outside of the `out.GetAgent().getName()` to make things more clear.

PR #206, opened 2024-02-02
action: use parenthesis and the correct sha

v1v:
Since it uses different on events, then it's required to set what's the sha commit to be used.

### DIFF
--- a/.github/workflows/bench-diff.yml
+++ b/.github/workflows/bench-diff.yml
@@ -13,7 +13,8 @@ jobs:
     # Support for branches, non-forked-prs or forked-prs with a label
     if: |
       contains(github.event.pull_request.labels.*.name, 'safe-to-test') ||
-      github.event_name != 'pull_request' || github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork == false
+      github.event_name != 'pull_request' ||
+      (github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork == false)
     permissions:
       checks: write
     steps:
--- a/.github/workflows/bench-diff.yml
+++ b/.github/workflows/bench-diff.yml
@@ -18,6 +18,11 @@ jobs:
     permissions:
       checks: write
     steps:
+    - name: Debug event types
+      run: |
+        echo "${{ github.event_name != 'pull_request' && github.sha || github.event.pull_request.head.sha }}"
+        echo "github.sha=${{ github.sha }}"
+        echo "github.event.pull_request.head.sha=${{ github.event.pull_request.head.sha }}"
     - uses: actions/checkout@v4
       with:
         fetch-depth: 0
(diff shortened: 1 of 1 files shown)


reakaleek (approved):

v1v:
I can only test this out once it's merged, at the moment it uses the commit from `main` :/

.github/workflows/bench-diff.yml
```diff
@@ -13,14 +13,21 @@ jobs:
     # Support for branches, non-forked-prs or forked-prs with a label
     if: |
       contains(github.event.pull_request.labels.*.name, 'safe-to-test') ||
-      github.event_name != 'pull_request' || github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork == false
+      github.event_name != 'pull_request' ||
+      (github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork == false)
```
reakaleek, line 17: I'm a bit confused about this part, actually.

According to the command forked-prs are supported.

Looking at the condition, forks are excluded?

.github/workflows/bench-diff.yml
```diff
@@ -29,7 +36,7 @@ jobs:
       id: benchdiff
       with:
         benchdiff_version: 0.9.1
-        status_sha: ${{ github.sha }}
+        status_sha: ${{ github.event_name != 'pull_request' && github.sha || github.event.pull_request.head.sha }}
```
reakaleek, line 39: If this runs on forks, it's a potential attack vector.

.github/workflows/bench-diff.yml
```diff
@@ -13,14 +13,21 @@ jobs:
...
+    - name: Debug event types
+      run: |
+        echo "${{ github.event_name != 'pull_request' && github.sha || github.event.pull_request.head.sha }}"
+        echo "github.sha=${{ github.sha }}"
+        echo "github.event.pull_request.head.sha=${{ github.event.pull_request.head.sha }}"
     - uses: actions/checkout@v4
       with:
         fetch-depth: 0
         persist-credentials: false
+        ref: ${{ github.event_name != 'pull_request' && github.sha || github.event.pull_request.head.sha }}
```
reakaleek, line 30: If this runs on forks, it's a potential attack vector.

.github/workflows/bench-diff.yml
v1v: It should run for the following events:

- push - if no a forked Pull Request
- label [safe-to-test](https://github.com/elastic/apm-data/labels/safe-to-test) - if a forked Pull Request

reakaleek: Ah, I read the comment wrong.

so it's 
- branches
- non-forked-prs
- forked-prs with the "safe-to-test" label.

I see

reakaleek (approved):

carsonip (approved):

PR #226, opened 2024-02-16
perf: do not pool on each iteration when processing the same batch

kruskall:
The processor is reading batchSize items at a time but on each iteration it retrieves and submit a batch to the pool. Now that we don't have the async code handling anymore we can just reuse the batch and only submit it to the pool once we're done reading the events.

### DIFF
--- a/input/elasticapm/processor.go
+++ b/input/elasticapm/processor.go
@@ -319,13 +319,12 @@ func (p *Processor) handleStream(
 	return readErr
 }
 
-// processBatch processes the batch and returns it to the pool after it's been processed.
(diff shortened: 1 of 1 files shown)


endorama:
> Now that we don't have the async code handling anymore

May you clarify this bit? Has this happened due to a recent code change?

endorama (approved):
One last question I have is how does the batch pooling works but I don't think is relevant for the review. To my understanding the code behaves like described in the PR description so :+1:

endorama:
@kruskall Thank you!

PR #243, opened 2024-03-21
feat: remove interfaces from protobuf models

kruskall:
slightly improve performance and avoid mapping errors now that we started indexing it with dynamic mapping.


input/elasticapm/docs/spec/v2/transaction.json
```diff
@@ -199,7 +199,13 @@
...
                 "null",
                 "object"
-              ]
+              ],
+              "additionalProperties": {
+                "type": [
+                  "null",
+                  "string"
+                ]
+              }
```
axw, line 208: Does this mean we'll ignore cookies that have non-string values? That would also be a breaking change, wouldn't it?

model/proto/http.proto
```diff
@@ -32,13 +33,14 @@ message HTTP {
 }
 
 message HTTPRequest {
+  reserved 4;
   google.protobuf.Value body = 1;
   repeated HTTPHeader headers = 2;
   repeated KeyValue env = 3;
-  repeated KeyValue cookies = 4;
```
axw, line 38: Should we keep and rename the old field, and convert to `map[string]string`? Otherwise we'll lose data?

input/elasticapm/docs/spec/v2/transaction.json
simitt: What is the current behaviour? Are non-string cookies accepted and then internally mapped to/handled as strings? I'd agree with @axw that this is a breaking change in behaviour then. 
Tbh I believe the switch to enforcing non-nil elements was borderline. With how big of a cost would backwards compatibility come here?

axw: By default we were not _indexing_ those fields (we were ingesting them!), but they are still present in the Elasticsearch documents. So a user could still inspect them in the document source, and they could have customised mappings to index them.

kyungeunni:
@kruskall Is there anything I can help to move this forward?

PR #246, opened 2024-03-27
Add support for providing host.id via intakeV2.

### LINKED ISSUE #187: Map the host.id opentelemetry attribute as top-level attribute
Our universal profiling solution relies on the `host.id` field, which is already part of ECS and the OTel Semconv.
In order to improve correlation between the profiling solution and APM, we should store `host.id` on span and transaction documents too.

APM (and OTel) agents will sent `host.id` as an opentelemetry attribute, we need to ensure that it is mapped as a top-level attribute instead of under `labels`. We will likely implement this as another "special case" in `apm-data`, so that we don't have to wait for the big change of storing all opentelemetry attributes at top-level by default.

JonasKunz:
Implementation for https://github.com/elastic/apm/pull/853, closes #187 .

We already support the [OpenTelemetry Resource Attribute host.id](https://opentelemetry.io/docs/specs/semconv/resource/host/) via the OTLP intake. Because this field is required for the universal profiling correlation, we need to be able to set that field via `IntakeV2`.
This PR therefore adds the optional `host_id` as new IntakeV2 metadata field.

### DIFF
--- a/input/elasticapm/internal/modeldecoder/v2/decoder.go
+++ b/input/elasticapm/internal/modeldecoder/v2/decoder.go
@@ -715,6 +715,12 @@ func mapToMetadataModel(from *metadata, out *modelpb.APMEvent) {
 		}
 		out.Host.Hostname = from.System.DetectedHostname.Val
 	}
+	if from.System.HostId.IsSet() {
+		if out.Host == nil {
+			out.Host = modelpb.HostFromVTPool()
+		}
+		out.Host.Id = from.System.HostId.Val
+	}
 	if !from.System.ConfiguredHostname.IsSet() && !from.System.DetectedHostname.IsSet() &&
 		from.System.DeprecatedHostname.IsSet() {
 		if out.Host == nil {
--- a/input/elasticapm/internal/modeldecoder/v2/metadata_test.go
+++ b/input/elasticapm/internal/modeldecoder/v2/metadata_test.go
@@ -319,9 +319,11 @@ func TestDecodeMapToMetadataModel(t *testing.T) {
 		input.System.ConfiguredHostname.Set("configured-host")
 		input.System.DetectedHostname.Set("detected-host")
 		input.System.DeprecatedHostname.Set("deprecated-host")
+		input.System.HostId.Set("host-id")
 		mapToMetadataModel(&input, &out)
 		assert.Equal(t, "configured-host", out.Host.Name)
 		assert.Equal(t, "detected-host", out.Host.Hostname)
+		assert.Equal(t, "host-id", out.Host.Id)
 		// no detected-host information
 		out = modelpb.APMEvent{}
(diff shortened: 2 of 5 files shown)


input/elasticapm/internal/modeldecoder/v2/model.go
```diff
@@ -594,6 +594,8 @@ type metadataSystem struct {
 	Kubernetes metadataSystemKubernetes `json:"kubernetes"`
 	// Platform name of the system platform the monitored service is running on.
 	Platform nullable.String `json:"platform" validate:"maxLength=1024"`
+	// The OpenTelemetry semantic conventions compliant "host.id" attribute, if available.
+	HostId nullable.String `json:"host_id" validate:"maxLength=1024"`
```
carsonip, line 598: the go specific way of variable naming
```suggestion
	HostID nullable.String `json:"host_id" validate:"maxLength=1024"`
```

input/elasticapm/internal/modeldecoder/v2/metadata_test.go
```diff
@@ -319,9 +319,11 @@ func TestDecodeMapToMetadataModel(t *testing.T) {
 		input.System.ConfiguredHostname.Set("configured-host")
 		input.System.DetectedHostname.Set("detected-host")
 		input.System.DeprecatedHostname.Set("deprecated-host")
+		input.System.HostId.Set("host-id")
```
carsonip, line 322:
```suggestion
		input.System.HostID.Set("host-id")
```

input/elasticapm/internal/modeldecoder/v2/decoder.go
```diff
@@ -715,6 +715,12 @@ func mapToMetadataModel(from *metadata, out *modelpb.APMEvent) {
 		}
 		out.Host.Hostname = from.System.DetectedHostname.Val
 	}
+	if from.System.HostId.IsSet() {
+		if out.Host == nil {
+			out.Host = modelpb.HostFromVTPool()
+		}
+		out.Host.Id = from.System.HostId.Val
```
carsonip, line 722:
```suggestion
		out.Host.Id = from.System.HostID.Val
```

input/elasticapm/internal/modeldecoder/v2/decoder.go
```diff
@@ -715,6 +715,12 @@ func mapToMetadataModel(from *metadata, out *modelpb.APMEvent) {
 		}
 		out.Host.Hostname = from.System.DetectedHostname.Val
 	}
+	if from.System.HostId.IsSet() {
```
carsonip, line 718:
```suggestion
	if from.System.HostID.IsSet() {
```

carsonip (dismissed):

JonasKunz:
@kruskall can you merge the PR for me? I don't have the permissions to do so

PR #258, opened 2024-04-10
Don't expose semaphore internals to the caller

vikmik:
The error message should not mention an internal implementation detail.


marclop (approved):
Thanks!

PR #263, opened 2024-04-15
Map updated HTTP server semantic conventions

### LINKED ISSUE #254: Map newer HTTP Server Semantic Conventions to URL
Since [v1.21.0](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.21.0/specification/trace/semantic_conventions/http.md#http-server-semantic-conventions) of the trace Semantic Conventions, new attributes are used:

- `net.host.name` -> `server.address`
- `net.host.port` -> `server.port`
- `http.target` -> `url.path` and `url.query`
- `http.scheme` -> `url.scheme`

We currently don't map all of these for the OTLP intake, resulting in a poor UI experience. It seems that only the scheme is coming through.

For example, the trace sample UI isn't able to display the URL of the request which started the transaction.

image

Should be closer to:

image

stevejgordon:
Adds support for updated semantic conventions for HTTP server spans.

Closes #254

### DIFF
--- a/input/otlp/traces.go
+++ b/input/otlp/traces.go
@@ -75,6 +75,9 @@ const (
 	attributeServerAddress              = "server.address"
 	attributeServerPort                 = "server.port"
 	attributeUrlFull                    = "url.full"
+	attributeUrlScheme                  = "url.scheme"
+	attributeUrlPath                    = "url.path"
+	attributeUrlQuery                   = "url.query"
 	attributeUserAgentOriginal          = "user_agent.original"
 	attributeDbElasticsearchClusterName = "db.elasticsearch.cluster.name"
 	attributeStackTrace                 = "code.stacktrace" // semconv 1.24 or later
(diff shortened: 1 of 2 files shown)


input/otlp/traces.go
```diff
@@ -450,11 +461,6 @@ func TranslateTransaction(
 			case semconv.AttributeRPCService:
 			case semconv.AttributeRPCMethod:
 
-			// URL
```
stevejgordon, line 453: Note for reviewer: The `url.full` attribute is not expected on spans for HTTP server requests, so I have removed this.

stevejgordon:
@simitt Here's an initial stab at supporting the newer attributes.

stevejgordon:
I rebased onto main via the UI, which seems to have lost the signing on the commit, which is annoying. @kruskall Are there any alternatives to recommitting locally (with the signing) and force pushing?

PR #268, opened 2024-04-25
Support events without domain attr

LikeTheSalad:
The `event.domain` attribute was removed from the semconv spec [here](https://github.com/open-telemetry/semantic-conventions/pull/473) and, the values that would otherwise be set into the `event.domain` attribute, must now be set as a namespace to the `event.name` value.

This PR is to make apm-data aware of this change while keeping backward compatibility, so that both implementations are properly identified as events.

### DIFF
--- a/input/otlp/logs.go
+++ b/input/otlp/logs.go
@@ -37,6 +37,7 @@ package otlp
 import (
 	"context"
 	"encoding/hex"
+	"strings"
 	"time"
 
 	"go.opentelemetry.io/collector/pdata/pcommon"
@@ -212,7 +213,7 @@ func (c *Consumer) convertLogRecord(
 		)
 	}
 
-	if eventDomain == "device" && eventName != "" {
+	if (eventDomain == "device" && eventName != "") || strings.HasPrefix(eventName, "device.") {
 		event.Event.Category = "device"
 		if eventName == "crash" {
 			if event.Error == nil {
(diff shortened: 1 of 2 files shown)


input/otlp/logs.go
```diff
@@ -212,16 +213,17 @@ func (c *Consumer) convertLogRecord(
 		)
 	}
 
-	if eventDomain == "device" && eventName != "" {
+	if (eventDomain == "device" && eventName != "") || strings.HasPrefix(eventName, "device.") {
 		event.Event.Category = "device"
-		if eventName == "crash" {
+		action := strings.TrimPrefix(eventName, "device.")
```
LikeTheSalad, line 218: The goal of trimming the `device.` prefix here is so that any existing verifications down the road for specific action names, such as `crash` for example, keep working as usual. However, I'm open to alternative approaches in case this change could cause issues that I might not be aware of.

input/otlp/logs.go
```diff
@@ -212,16 +213,17 @@ func (c *Consumer) convertLogRecord(
 		)
 	}
 
-	if eventDomain == "device" && eventName != "" {
+	if (eventDomain == "device" && eventName != "") || strings.HasPrefix(eventName, "device.") {
```
bryce-b, line 216: could you add a comment here that notes the version of the otel spec this is accommodating?

bryce-b (approved):

input/otlp/logs.go
LikeTheSalad: Sure, it's done now.

PR #277, opened 2024-05-09
Remap OTel hostmetrics to Elastic metrics

lahsivjar:
The PR uses the library [opentelemetry-lib](https://github.com/elastic/opentelemetry-lib/) to remap metrics produced by [hostmetricsreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/README.md) to Elastic metrics that could power the curated UIs.

<details>
<summary>Example OTel collector configuration required to get the remapping working</summary>

```yaml
receivers:
  hostmetrics:
    collection_interval: 10s
    scrapers:
      load:
      cpu:
        metrics:
          system.cpu.utilization:
            enabled: true
          system.cpu.logical.count:
            enabled: true
      memory:
        metrics:
          system.memory.utilization:
            enabled: true
      network:
      process:
        metrics:
          process.threads:
            enabled: true
          process.open_file_descriptors:
            enabled: true
          process.memory.utilization:
            enabled: true
          process.disk.operations:
            enabled: true
      processes:

exporters:
  otlphttp:
    endpoint: http://192.168.0.44:8200
    tls:
      insecure: true
  debug:
    verbosity: detailed

processors:
  resourcedetection/system:
    detectors: ["system"]
    system:
      hostname_sources: ["os"]

service:
  pipelines:
    metrics:
      receivers: [hostmetrics]
      processors: [resourcedetection/system]
      exporters: [otlphttp, debug]
```
</details>

## Related issues

- https://github.com/elastic/opentelemetry-dev/issues/217
- https://github.com/elastic/opentelemetry-dev/issues/180

### DIFF
--- a/input/otlp/metrics.go
+++ b/input/otlp/metrics.go
@@ -152,9 +152,7 @@ func (c *Consumer) handleScopeMetrics(
 		for _, r := range c.remappers {
 			r.Remap(in, remappedMetrics, resource)
 		}
-		// Add each remapped metric to the metricset. Code assumes that
-		// a single datapoint is added for each metric by the library.
-		*remainingDataPoints += int64(remappedMetrics.Len())
+		*remainingDataPoints += int64(dataPointsCount(remappedMetrics))
 		*remainingMetrics += int64(remappedMetrics.Len())
 		for i := 0; i < remappedMetrics.Len(); i++ {
 			c.addMetric(remappedMetrics.At(i), ms, remainingDataPoints, remainingMetrics)
@@ -396,4 +394,23 @@ func (ms metricsets) upsertOne(timestamp time.Time, attributes pcommon.Map, samp
 		ms[key] = m
 	}
 	m.samples[sample.Name] = sample
+}
+
+func dataPointsCount(ms pmetric.MetricSlice) (count int) {
+	for i := 0; i < ms.Len(); i++ {
+		m := ms.At(i)
+		switch m.Type() {
+		case pmetric.MetricTypeGauge:
+			count += m.Gauge().DataPoints().Len()
+		case pmetric.MetricTypeSum:
+			count += m.Sum().DataPoints().Len()
+		case pmetric.MetricTypeHistogram:
+			count += m.Histogram().DataPoints().Len()
+		case pmetric.MetricTypeExponentialHistogram:
+			count += m.ExponentialHistogram().DataPoints().Len()
+		case pmetric.MetricTypeSummary:
+			count += m.Summary().DataPoints().Len()
+		}
+	}
+	return
 }
--- a/input/otlp/metrics.go
+++ b/input/otlp/metrics.go
@@ -146,7 +146,6 @@ func (c *Consumer) handleScopeMetrics(
 	}
 	// Handle remapping if any. Remapped metrics will be added to a new
 	// metric slice and then processed as any other metric in the scope.
-	// TODO (lahsivjar): Possible to approximate capacity of the slice?
 	if len(c.remappers) > 0 {
 		remappedMetrics := pmetric.NewMetricSlice()
 		for _, r := range c.remappers {
--- a/input/otlp/metrics.go
+++ b/input/otlp/metrics.go
@@ -75,7 +75,7 @@ func (c *Consumer) ConsumeMetricsWithResult(ctx context.Context, metrics pmetric
 	remainingMetrics := totalMetrics
 	receiveTimestamp := time.Now()
 	c.config.Logger.Debug("consuming metrics", zap.Stringer("metrics", metricsStringer(metrics)))
-	batch := c.convertMetrics(metrics, receiveTimestamp, &remainingDataPoints, &remainingMetrics)
+	batch := c.handleMetrics(metrics, receiveTimestamp, &remainingDataPoints, &remainingMetrics)
 	if remainingMetrics > 0 {
 		// Some metrics remained after conversion, meaning that they were dropped.
 		atomic.AddInt64(&c.stats.unsupportedMetricsDropped, remainingMetrics)
@@ -93,7 +93,7 @@ func (c *Consumer) ConsumeMetricsWithResult(ctx context.Context, metrics pmetric
 	}, nil
 }
 
-func (c *Consumer) convertMetrics(
+func (c *Consumer) handleMetrics(
 	metrics pmetric.Metrics,
 	receiveTimestamp time.Time,
 	remainingDataPoints, remainingMetrics *int64,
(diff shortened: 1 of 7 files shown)


model/modelprocessor/datastream.go
```diff
@@ -141,10 +141,18 @@ func metricsetDataset(event *modelpb.APMEvent) string {
...
+		if label, ok := event.Labels["event.provider"]; ok && label != nil {
+			internal = !(label.Value == "hostmetrics")
+		}
+
+		if internal {
+			for _, s := range event.Metricset.Samples {
+				if !IsInternalMetricName(s.Name) {
+					internal = false
+					break
+				}
```
lahsivjar, line 155: [**For reviewers**] This code is required to make sure that the remapped OTel metrics end up in the app datastream instead of the internal datastream.

input/otlp/metrics.go
```diff
@@ -125,23 +125,42 @@ func (c *Consumer) convertResourceMetrics(
...
+	// Handle remapping if any. Remapped metrics will be added to a new
+	// metric slice and then processed as any other metric in the scope.
+	// TODO (lahsivjar): Possible to approximate capacity of the slice?
+	if len(c.remappers) > 0 {
+		remappedMetrics := pmetric.NewMetricSlice()
+		for _, r := range c.remappers {
+			r.Remap(in, remappedMetrics, resource)
+		}
+		// Add each remapped metric to the metricset. Code assumes that
+		// a single datapoint is added for each metric by the library.
```
endorama, line 156: (blocking) Under which condition does this assumption hold true? I'm worries this could silently drop data points and investigating it would be quite painful.

input/otlp/metrics.go
```diff
@@ -125,23 +125,42 @@ func (c *Consumer) convertResourceMetrics(
...
 ) {
 	ms := make(metricsets)
+	// Add the original otel metrics to the metricset.
 	otelMetrics := in.Metrics()
 	for i := 0; i < otelMetrics.Len(); i++ {
 		c.addMetric(otelMetrics.At(i), ms, remainingDataPoints, remainingMetrics)
 	}
+	// Handle remapping if any. Remapped metrics will be added to a new
+	// metric slice and then processed as any other metric in the scope.
+	// TODO (lahsivjar): Possible to approximate capacity of the slice?
```
endorama, line 149: Is it possible to count the metrics in `in`? If yes would multiply them by `c.remappers` produce a decently accurate figure?

model/modelprocessor/datastream.go
```diff
@@ -141,10 +141,18 @@ func metricsetDataset(event *modelpb.APMEvent) string {
 		// metrics that we don't already know about; otherwise they will end
 		// up creating service-specific data streams.
 		internal := true
-		for _, s := range event.Metricset.Samples {
-			if !IsInternalMetricName(s.Name) {
-				internal = false
-				break
+
+		// set internal to false for metrics translated using OTel remappers.
```
endorama, line 145: Do we plan to allow to remap further metrics in the future? I'm worried this code could be difficult to change if we want to remap different metrics in the future. Disregard this comment if this could be easily solved just looping over `event` attributes to achieve it.

endorama (commented):
Left a blocking comment and some minor non-blocking discussion points.

input/otlp/metrics.go
lahsivjar: > Is it possible to count the metrics in in?

Unfortunately no, the remapping library has full control over which metrics will be remapped and thus the count of the remapped slice. We can get an upper bound but in most cases, it would be too big. I was thinking of adding a method in the remappers to get an approximate size (which each remapper can give) but maintaining such code would be a PITA as adding any new metric would require to update the count. So, TL;DR I don't have a good approach yet.

input/otlp/metrics.go
lahsivjar: Hmm, fair point. The way the remapping library is written this will hold for all cases, however, it might come to bite us in future if the implementation changes. Let me fix this.

model/modelprocessor/datastream.go
lahsivjar: Sorry, I didn't get this point.

> Do we plan to allow to remap further metrics in the future?

Yes we will be having more remapped metrics in the future.

> I'm worried this code could be difficult to change if we want to remap different metrics in the future. Disregard this comment if this could be easily solved just looping over event attributes to achieve it.

All remapped metrics should go to apm.app datastream so I don't see why we would require to loop over anything. Let me know if I am missing the point here.

input/otlp/metrics.go
lahsivjar: Done!

input/otlp/metrics.go
axw: Do we expect a high volume of hits? I wouldn't think so - so it probably doesn't matter much.

model/modelprocessor/datastream.go
axw: IIANM we'll also be remapping k8s metrics? In which case they would come from some other receiver IIANM. I wonder if we should just check for the `event.provider` key, and treat them as app metrics if the key exists regardless of the value.

axw (commented):
Code LGTM, just wondering about the `event.provider` check. I think we should come to an agreement on that before merging.

model/modelprocessor/datastream.go
endorama: > Maybe you mean what should we do in case the remapping has a different way to identify it?

@lahsivjar yes indeed, sorry for the confusion (my comment would have been better of on the next `if` when `event.provider` is checked).

input/otlp/metrics.go
lahsivjar: > Do we expect a high volume of hits? I wouldn't think so - so it probably doesn't matter much.

Agreed.

model/modelprocessor/datastream.go
lahsivjar: The key for identifying remapped metrics was changed to `event.module: elastic/opentelemetry-lib` via https://github.com/elastic/opentelemetry-lib/pull/18. I have updated the code as per the changes. We will always have this key for all remapped metrics. Let me know if there are any other concerns regarding this.

endorama (approved):
Looks good to me!

input/otlp/metrics.go
```diff
@@ -125,23 +125,40 @@ func (c *Consumer) convertResourceMetrics(
...
```
axw, line 149:
```suggestion
```

axw (approved):

axw (approved):

PR #309, opened 2024-07-03
Handle system.process.* for OTel translated metrics

lahsivjar:
`system.process.*` fields are required on the process metrics to make the UI work. These are non-ecs fields but must haves for the Processes tab in the Host metrics UI to work.

All these fields are processed as data point attributes added to the remapped OTel metrics due to UI constraints (explained in code comments).

Related to:
- https://github.com/elastic/opentelemetry-lib/pull/34
- https://github.com/elastic/opentelemetry-dev/issues/217
- https://github.com/elastic/opentelemetry-dev/issues/180

To be merged after https://github.com/elastic/opentelemetry-lib/pull/35


input/otlp/metrics.go
```diff
@@ -186,6 +186,55 @@ func (c *Consumer) handleScopeMetrics(
...
+					}
+					event.System.Process.Cpu.StartTime = v.Str()
+
+				// `system.process.cmdline` is same as the ECS field `process.command_line`
+				// however, Kibana curated UIs requires this field to work. In addition,
+				// the current Kibana code will not work if this field is added to documents
+				// with `system.process.memory.rss.pct` and other metrics required in the
+				// Processes tab of the Kibana hostmetrics UI. Due to this, we have to process
+				// the datapoint field added by the opentelemetry-lib instead of directly
+				// processing the OTel semconv resource attribute `process.command_line`.
```
lahsivjar, line 213: @felixbarny @axw This means that we would not be able to remove the attributes added at https://github.com/elastic/opentelemetry-lib/blob/54a9569b27e3f11edc0d660dc79ac865256f3cd3/remappers/hostmetrics/process.go#L271-L295 and rely on the translations performed in APM/esexporter without UI fixing such hard dependencies. One such case I have explained [here](https://github.com/elastic/observability-dev/issues/3704#issuecomment-2207351332) (elastic only link).

axw (approved):

input/otlp/metrics.go
```diff
@@ -186,6 +186,55 @@ func (c *Consumer) handleScopeMetrics(
...
+				case "system.process.cmdline":
+					if event.System == nil {
+						event.System = modelpb.SystemFromVTPool()
+					}
+					if event.System.Process == nil {
+						event.System.Process = modelpb.SystemProcessFromVTPool()
+					}
+					event.System.Process.Cmdline = v.Str()
```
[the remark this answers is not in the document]
lahsivjar: Hmm, good point. We should I think truncate others too just like what we do in the metadata. On it.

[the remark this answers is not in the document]
lahsivjar: Done, only the cmdline required truncating.

input/otlp/metrics_test.go
```diff
@@ -904,17 +915,61 @@ func TestConsumeMetricsWithOTelRemapper(t *testing.T) {
...
+				Name:     "unknown",
+				Language: &modelpb.Language{Name: "unknown"},
+			},
+			Agent:     &modelpb.Agent{Name: "otlp", Version: "unknown"},
+			Timestamp: modelpb.FromTime(ts),
+			Metricset: &modelpb.Metricset{
+				Name: "app",
+				Samples: []*modelpb.MetricsetSample{
 					{
-						Name: "system.load.5",
```
carsonip, line 912: Question: Why is this here in the first place with no value, and why is it not present now after the PR?

carsonip (approved):

input/otlp/metrics_test.go
lahsivjar: It is due to how otel lib behaves. There was a fix in v0.6.0 but that had to reverted in v0.6.1 so these lines need to be added back. I have added a comment explaining why it is there pasting below for reference:

```
				// TODO (lahsivjar): Opentelemetry lib currently adds all metrics, even
				// though the source metrics is not present. Due to this all the metrics
				// need to be asserted making the tests brittle. This should be fixed by
				// https://github.com/elastic/opentelemetry-lib/issues/6
```

carsonip (approved):

PR #312, opened 2024-07-04
Fix flaky TestConsumeMetricsWithOTelRemapper

lahsivjar:
The flakiness in the test `TestConsumeMetricsWithOTelRemapper` was introduced in https://github.com/elastic/apm-data/pull/309 when we added another scope in the metric.

The root cause of the flakiness is the function `eventsMatch`. This function sorts the APMEvents by converting it into strings and later compares them at which point it uses `protocmp.SortRepeated` for metricset samples. The metricsets are created using maps ([ref](https://github.com/elastic/apm-data/blob/e5d0335650003df5fac79cd8485cb775bac221b5/input/otlp/metrics.go#L403)) which makes the order of metricsets undefined. While the tests use sorting on the metricset samples to get around this we still end up in an indeterministic state since the first sort doesn't sort on metricset samples ([ref](https://github.com/elastic/apm-data/blob/e5d0335650003df5fac79cd8485cb775bac221b5/input/otlp/metrics_test.go#L1087-L1092)).

We can fix the `eventsMatch` function by first sorting the samples in each `APMEvent` and then sorting the `APMEvent`'s themselves and this is what the PR does. In addition the PR also refactors the test `TestConsumeMetricsWithOTelRemapper` which would avoid the special issue anyway.

### DIFF
--- a/input/otlp/metrics_test.go
+++ b/input/otlp/metrics_test.go
@@ -855,200 +855,222 @@ func TestConsumeMetricsDataStream(t *testing.T) {
 // code but does not test the correctness of the remapping libarary or if all
 // the metrics are remapped as it is within the scope of the remapping library.
 func TestConsumeMetricsWithOTelRemapper(t *testing.T) {
-	metrics := pmetric.NewMetrics()
-	rm := metrics.ResourceMetrics().AppendEmpty()
-	sm := rm.ScopeMetrics().AppendEmpty()
-
 	ts := time.Now().UTC()
 	startTs := ts.Add(-time.Hour)
 
-	// Configure scope for hostmetrics receiver
-	sm.Scope().SetName("otelcol/hostmetricsreceiver/load")
-
-	// Add a datapoint for a metric produced by hostmetrics receiver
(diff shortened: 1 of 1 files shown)


input/otlp/metrics_test.go
```diff
@@ -855,200 +855,222 @@ func TestConsumeMetricsDataStream(t *testing.T) {
...
-
-	// process metric
-	sm = rm.ScopeMetrics().AppendEmpty()
-	sm.Scope().SetName("otelcol/hostmetricsreceiver/process")
-	metric = sm.Metrics().AppendEmpty()
-	metric.SetName("process.memory.usage")
-	dp = metric.SetEmptySum().DataPoints().AppendEmpty()
-	dp.SetTimestamp(pcommon.NewTimestampFromTime(ts))
-	dp.SetStartTimestamp(pcommon.NewTimestampFromTime(startTs))
-	dp.SetIntValue(1024)
```
lahsivjar, line 883: [**For reviewers**] There should be no change in the tests. The tests just divides the load and the process scope into two separate tests.

Issue #326, opened 2024-07-22
Panic in stacktrace mapping


lahsivjar:
```
{"log.level":"error","@timestamp":"2024-07-22T15:24:51.682+0100","log.logger":"request","log.origin":{"function":"github.com/elastic/apm-server/internal/beater/api.apmMiddleware.LogMiddleware.func1.1","file.name":"middleware/log_middleware.go","file.line":59},"message":"internal error","service.name":"apm-server","url.original":"/intake/v2/events","http.request.method":"POST","user_agent.original":"Go-http-client/1.1","source.address":"127.0.0.1","http.request.id":"a390e2cf-1bbc-4485-9251-4f082c5b2cbe","event.duration":449042,"http.request.body.bytes":4096,"http.response.status_code":500,"error.message":"internal error: runtime error: invalid memory address or nil pointer dereference","error.stack_trace":"goroutine 180 [running]:
	runtime/debug.Stack()
		runtime/debug/stack.go:24 +0x64
	github.com/elastic/apm-server/internal/beater/api.apmMiddleware.RecoverPanicMiddleware.func2.1.1()
		github.com/elastic/apm-server/internal/beater/middleware/recover_panic_middleware.go:55 +0x134
	panic({0x106a3dd20?, 0x108025640?})
		runtime/panic.go:770 +0x124
	github.com/elastic/apm-data/input/elasticapm/internal/modeldecoder/v2.mapToStracktraceModel({0x1400f88a008?, 0x1a, 0x140028c2120?}, {0x14018f18f00, 0x1a, 0x140198f5ada?})
		github.com/elastic/apm-data@v1.9.0/input/elasticapm/internal/modeldecoder/v2/decoder.go:1320 +0x80
	github.com/elastic/apm-data/input/elasticapm/internal/modeldecoder/v2.mapToSpanModel(0x140115ec608, 0x1401860dce0)
		github.com/elastic/apm-data@v1.9.0/input/elasticapm/internal/modeldecoder/v2/decoder.go:1282 +0x13f0
	github.com/elastic/apm-data/input/elasticapm/internal/modeldecoder/v2.DecodeNestedSpan({0x106d5b2e0, 0x14018c351e0}, 0x140000bceb8, 0x14018c62c18)
		github.com/elastic/apm-data@v1.9.0/input/elasticapm/internal/modeldecoder/v2/decoder.go:218 +0x118
[...]

PR #327, opened 2024-07-23
Fix NPE due to reslice only initializing new slice elements

lahsivjar:
Fixes https://github.com/elastic/apm-data/issues/326

Run the test case in https://github.com/elastic/apm-data/issues/326#issuecomment-2245636455 to understand the root issue. Note that the test case is not added here since it is too specific for reproducing the bug and, IMO, will not be useful as a test for long-term maintenance.

### DIFF
--- a/input/elasticapm/internal/modeldecoder/modeldecoderutil/slice.go
+++ b/input/elasticapm/internal/modeldecoder/modeldecoderutil/slice.go
@@ -24,7 +24,7 @@ import (
 // Reslice returns a slice with length n. It will reallocate the
 // slice if the capacity is not enough.
 func Reslice[Slice ~[]model, model any](slice Slice, n int) Slice {
-	if n-cap(slice) > 0 {
+	if n > cap(slice) {
 		slice = slices.Grow(slice, n-len(slice))
 	}
 	return slice[:n]
--- a/input/elasticapm/internal/modeldecoder/modeldecoderutil/slice.go
+++ b/input/elasticapm/internal/modeldecoder/modeldecoderutil/slice.go
@@ -30,13 +30,16 @@ func Reslice[Slice ~[]model, model any](slice Slice, n int) Slice {
 	return slice[:n]
 }
 
-// PopulateNil populates all the nil elements of a pointer slice
-// with the value returned from the passed newFn. It returns the
-// slice to allow chaining.
-func PopulateNil[Slice ~[]*model, model any](slice Slice, newFn func() *model) Slice {
-	for i := 0; i < len(slice); i++ {
-		if slice[i] == nil {
-			slice[i] = newFn()
+// ResliceAndPopulateNil ensures a slice of pointers has atleast
+// capacity for n elements and populates any non-nil elements
+// in the resulting slice with the value returned from newFn.
+func ResliceAndPopulateNil[Slice ~[]*model, model any](slice Slice, n int, newFn func() *model) Slice {
+	slice = Reslice(slice, n)
+	if newFn != nil {
+		for i := 0; i < len(slice); i++ {
+			if slice[i] == nil {
+				slice[i] = newFn()
+			}
 		}
 	}
 	return slice
--- a/input/elasticapm/internal/modeldecoder/modeldecoderutil/slice.go
+++ b/input/elasticapm/internal/modeldecoder/modeldecoderutil/slice.go
@@ -17,28 +17,27 @@
 
 package modeldecoderutil
 
-// Reslice increases the slice's capacity, if necessary, to match n.
-// If specified, the newFn function is used to create the elements
-// to populate the additional space appended to the slice.
-func Reslice[Slice ~[]model, model any](slice Slice, n int, newFn func() model) Slice {
-	if diff := n - cap(slice); diff > 0 {
-		// start of the extra space
-		idx := cap(slice)
+import (
+	"slices"
+)
 
-		// Grow the slice
-		// Note: append gives no guarantee on the capacity of the resulting slice
-		// and might overallocate as long as there's enough space for n elements.
-		slice = append([]model(slice)[:cap(slice)], make([]model, diff)...)
-		if newFn != nil {
-			// extend the slice to its capacity
-			slice = slice[:cap(slice)]
+// Reslice returns a slice with length n. It will reallocate the
+// slice if the capacity is not enough.
+func Reslice[Slice ~[]model, model any](slice Slice, n int) Slice {
+	if n-cap(slice) > 0 {
+		slice = slices.Grow(slice, n-len(slice))
+	}
+	return slice[:n]
+}
 
-			// populate the extra space
-			for ; idx < len(slice); idx++ {
-				slice[idx] = newFn()
-			}
+// PopulateNil populates all the nil elements of a pointer slice
+// with the value returned from the passed newFn. It returns the
+// slice to allow chaining.
+func PopulateNil[Slice ~[]*model, model any](slice Slice, newFn func() *model) Slice {
+	for i := 0; i < len(slice); i++ {
+		if slice[i] == nil {
+			slice[i] = newFn()
 		}
 	}
-
-	return slice[:n]
+	return slice
 }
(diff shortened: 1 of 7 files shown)


input/elasticapm/internal/modeldecoder/modeldecoderutil/slice.go
```diff
@@ -17,28 +17,27 @@
...
-		// Grow the slice
-		// Note: append gives no guarantee on the capacity of the resulting slice
-		// and might overallocate as long as there's enough space for n elements.
-		slice = append([]model(slice)[:cap(slice)], make([]model, diff)...)
-		if newFn != nil {
-			// extend the slice to its capacity
-			slice = slice[:cap(slice)]
+// Reslice returns a slice with length n. It will reallocate the
+// slice if the capacity is not enough.
+func Reslice[Slice ~[]model, model any](slice Slice, n int) Slice {
```
lahsivjar, line 26: [**For reviewers**] I am not a fan of having such util methods but I didn't want to be too disruptive here. Let me know if anyone has a better idea.

carsonip: question: can we maintain the original signature of the Reslice function to reduce the changes required? Apparently we never use Reslice and PopulateNil separately, keeping it down to 1 function means we'll not risk missing the call to PopulateNil. WDYT?

input/elasticapm/internal/modeldecoder/modeldecoderutil/slice.go
```diff
@@ -17,28 +17,27 @@
...
+	if n-cap(slice) > 0 {
```
carsonip, line 27: Is this redundant? IIRC this check already exists in slices.Grow.

lahsivjar: Grow will panic on negative vals and without check we can endup in negative range.

input/elasticapm/internal/modeldecoder/modeldecoderutil/slice.go
lahsivjar: I don't think there is a good optimized way to do that since Reslice is called for both pointers and strings and checking for nil/empty will not be that straightforward with generics. Additionally, populate nil functionality doesn't fit properly in reslice IMO... it is too confusing with what it is supposed to do.

carsonip (commented):
nit: As PopulateNil allows chaining, wdyt about collapsing all the calls to e.g. `	out = PopulateNil(Reslice(out, len(h)), modelpb.HTTPHeaderFromVTPool)`? I think it looks slightly better than calling it inside `mapTo*` e.g.
```
out.Stacktrace = modeldecoderutil.Reslice(out.Stacktrace, len(from.Stacktrace))
mapToStracktraceModel(from.Stacktrace, modeldecoderutil.PopulateNil(out.Stacktrace, modelpb.StacktraceFrameFromVTPool))
```

input/elasticapm/internal/modeldecoder/modeldecoderutil/slice.go
lahsivjar: An alternative way could be to have one method `ResliceAndPopulateNil` for pointers and another `Reslice` as a generic... that might turn out to be a bit better. I will do that in a bit today or tomorrow and see how it turns out.

carsonip: Oh, now I understand your point about pointers and strings, and how the generic types don't match. I'll approve this PR as is, and feel free to merge it if you can't figure out a better way.

carsonip (dismissed):

lahsivjar:
@kruskall Yup, you are right they are doing the same thing. Do you want to take that forward instead?

input/elasticapm/internal/modeldecoder/modeldecoderutil/slice.go
```diff
@@ -17,28 +17,30 @@
...
```
carsonip, line 27: nit: why `n-cap(slice)` instead of `n-len(slice)`, if its sole purpose is to avoid going into negative? I suppose the result is the same, but personally I prefer either `if n-len(slice) > 0` or `if n > cap(slice)` for clarity

carsonip (approved):

input/elasticapm/internal/modeldecoder/modeldecoderutil/slice.go
lahsivjar: >why n-cap(slice) instead of n-len(slice), if its sole purpose is to avoid going into negative? 

If we have the capacity there is no point in even calling Grow.

lahsivjar: > but personally I prefer either if n-len(slice) > 0 or if n > cap(slice) for clarity

Not sure how it will be clearer honestly but I have no strong inclination so let me update with `n > cap(slice)`

carsonip (approved):

carsonip:
✅ test-plan-ok

Adding the following test (supplied in [original issue](https://github.com/elastic/apm-data/issues/326#issuecomment-2245636455)) to  `input/elasticapm/processor_test.go`:

```go

func TestBadReslice(t *testing.T) {
	metadata := `{"metadata": {"service": {"name": "1234_service-12a3","node": {"configured_name": "node-123"},"version": "5.1.3","environment": "staging","language": {"name": "ecmascript","version": "8"},"runtime": {"name": "node","version": "8.0.0"},"framework": {"name": "Express","version": "1.2.3"},"agent": {"name": "elastic-node","version": "3.14.0"}},"user": {"id": "123user", "username": "bar", "email": "bar@user.com"}, "labels": {"tag0": null, "tag1": "one", "tag2": 2}, "process": {"pid": 1234,"ppid": 6789,"title": "node","argv": ["node","server.js"]},"system": {"hostname": "prod1.example.com","architecture": "x64","platform": "darwin", "container": {"id": "container-id"}, "kubernetes": {"namespace": "namespace1", "pod": {"uid": "pod-uid", "name": "pod-name"}, "node": {"name": "node-name"}}},"cloud":{"account":{"id":"account_id","name":"account_name"},"availability_zone":"cloud_availability_zone","instance":{"id":"instance_id","name":"instance_name"},"machine":{"type":"machine_type"},"project":{"id":"project_id","name":"project_name"},"provider":"cloud_provider","region":"cloud_region","service":{"name":"lambda"}}}}`
	errstacktrace := `{"error": {"id": "cdefab0123456789", "trace_id": null, "timestamp": 1533826745999000,"exception": {"message": "Cannot read property 'baz' no defined", "stacktrace": [{"classname": "cl1", "abs_path": "abs_path1"}, {"classname": "cl2", "abs_path": "abs_path2"}, {"classname": "cl3", "abs_path": "abs_path3"}, {"classname": "cl4", "abs_path": "abs_path4"}, {"classname": "cl5", "abs_path": "abs_path5"}]}}}`

	batchProcessor := modelpb.ProcessBatchFunc(func(ctx context.Context, batch *modelpb.Batch) error {
		return nil
	})
	p := NewProcessor(Config{
		MaxEventSize: 100 * 1024,
		Semaphore:    semaphore.NewWeighted(1),
	})

	// Below, we try to emulate a case where all of the capacity of Error#Exception#Stacktrace slice
	// is not populated with non-nil values. In practical cases, this could happen with OTLP handling
	// of exceptions: https://github.com/elastic/apm-data/blob/cee5ac3fc5f2ee66f156133992882cae758e2b66/input/otlp/exceptions.go#L165
	event1 := modelpb.APMEventFromVTPool()
	event1.Error = modelpb.ErrorFromVTPool()
	event1.Error.Exception = modelpb.ExceptionFromVTPool()
	for i := 0; i < 3; i++ {
		// Add 3 so that reallocation makes the capacity 4
		// this will cause the 4th element to be nil
		event1.Error.Exception.Stacktrace = append(event1.Error.Exception.Stacktrace, modelpb.StacktraceFrameFromVTPool())
	}
	require.Equal(t, 4, cap(event1.Error.Exception.Stacktrace))

	event1.ReturnToVTPool()
	event2 := modelpb.APMEventFromVTPool()
	payload2 := strings.Join([]string{metadata, errstacktrace}, "\n")
	err := p.HandleStream(
		context.Background(), event2,
		strings.NewReader(payload2), 10, batchProcessor,
		&Result{},
	)
	require.NoError(t, err)
}
```

Confirmed that it fails in v1.9.0 and passes in v1.9.1.

On a side note, I cannot wrap my head around how this would manifest practically. There is no code in apm-server other than otlp code that would possibly insert nil to stacktrace slice, but otlp APMEvents are not pooled. But investigation into that is a separate matter. This PR is test-plan-ok as it defends against bad pooled slices containing nils.

PR #341, opened 2024-08-09
Fix OTEL Exception Message Propagation

rubvs:
Fix on [Issue 13603](https://github.com/elastic/apm-server/issues/13603)

The issue is reproduced by:

- Running `apm-server` locally.
- Sending OTLP data via the Go client `sendotlp`.
    - Some changes was made to `apm-server/systemtest/cmd/sendotlp`
    - I'll open a PR with this in `apm-server` once this PR is approved
- View the document in Table form in Kibana -> Analytics -> Discover

incorrect

The issue is fixed by:

- Applying the changes from `apm-data` to `apm-server` 
    - `cd apm-server`
    - `go mod edit -replace=github.com/elastic/apm-data=../apm-data`
    - `make update`
- Redo steps as depicted above.

correct

### DIFF
--- a/model/modelprocessor/errormessage.go
+++ b/model/modelprocessor/errormessage.go
@@ -33,6 +33,9 @@ func (s SetErrorMessage) ProcessBatch(ctx context.Context, b *modelpb.Batch) err
 		event := (*b)[i]
 		if event.Error != nil {
 			event.Error.Message = s.setErrorMessage(event)
+			if event.GetMessage() == "" {
+				event.Message = event.Error.Message
+			}
 		}
 	}
 	return nil
--- a/model/modelprocessor/errormessage.go
+++ b/model/modelprocessor/errormessage.go
@@ -32,9 +32,8 @@ func (s SetErrorMessage) ProcessBatch(ctx context.Context, b *modelpb.Batch) err
 	for i := range *b {
 		event := (*b)[i]
 		if event.Error != nil {
-			event.Error.Message = s.setErrorMessage(event)
 			if event.GetMessage() == "" {
-				event.Message = event.Error.Message
+				event.Message = s.setErrorMessage(event)
 			}
 		}
 	}
--- a/model/modelprocessor/errormessage.go
+++ b/model/modelprocessor/errormessage.go
@@ -32,7 +32,7 @@ func (s SetErrorMessage) ProcessBatch(ctx context.Context, b *modelpb.Batch) err
 	for i := range *b {
 		event := (*b)[i]
 		if event.Error != nil {
-			event.Message = s.setErrorMessage(event)
+			event.Error.Message = s.setErrorMessage(event)
 		}
 	}
 	return nil
--- a/model/modelprocessor/errormessage_test.go
+++ b/model/modelprocessor/errormessage_test.go
@@ -27,75 +27,27 @@ import (
 	"github.com/elastic/apm-data/model/modelprocessor"
 )
 
-// Copy the error message to the event message if event message is empty.
-func TestSetErrorMessage(t *testing.T) {
-	tests := []struct {
-		input    *modelpb.Error
-		expected string
-	}{{
-		input:    &modelpb.Error{},
-		expected: "",
-	}, {
-		input:    &modelpb.Error{Log: &modelpb.ErrorLog{Message: "log_message"}},
-		expected: "log_message",
-	}, {
-		input:    &modelpb.Error{Exception: &modelpb.Exception{Message: "exception_message"}},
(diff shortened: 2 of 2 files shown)


carsonip (commented):
https://github.com/elastic/apm-data/pull/341/files#diff-a83af8750cbc259cb07425b435f9f248022e844152a698eb5aca74dcad7b10a0R26
> // SetErrorMessage is a modelpb.BatchProcessor that sets the APMEvent.Message
// field for error events.

Have you figured out why it was implemented like that in the first place? If it is a bug, do you mind updating the comment as well?

rubvs:
> https://github.com/elastic/apm-data/pull/341/files#diff-a83af8750cbc259cb07425b435f9f248022e844152a698eb5aca74dcad7b10a0R26
> 
> > // SetErrorMessage is a modelpb.BatchProcessor that sets the APMEvent.Message
> > // field for error events.
> 
> Have you figured out why it was implemented like that in the first place? If it is a bug, do you mind updating the comment as well?

@kruskall if I'm not mistaken, you added this comment, can you maybe elaborate on this please?

axw:
Thanks for the ping @kruskall 

I think `SetErrorMessage` predates our support for OTLP logs, and was implemented only with exception span events in mind. In that case `message` will not be set at all, prior to `SetErrorMessage` copying the value across.

The point of setting the `message` field was to ensure the exceptions show up in the Logs app. Not sure if this is necessary any more, but it was back then.

Perhaps the simplest thing to do would be to change `SetErrorMessage` to only set `message` if it's empty.

model/modelprocessor/errormessage.go
```diff
@@ -32,7 +32,10 @@ func (s SetErrorMessage) ProcessBatch(ctx context.Context, b *modelpb.Batch) err
 	for i := range *b {
 		event := (*b)[i]
 		if event.Error != nil {
-			event.Message = s.setErrorMessage(event)
+			event.Error.Message = s.setErrorMessage(event)
```
carsonip, line 35: > Perhaps the simplest thing to do would be to change SetErrorMessage to only set message if it's empty.

My understanding of @axw 's comment is that we don't touch event.Error.Message, and only set event.Message if it is empty.

model/modelprocessor/errormessage_test.go
```diff
@@ -27,39 +27,90 @@ import (
...
 		err := processor.ProcessBatch(context.Background(), &batch)
 		assert.NoError(t, err)
-		assert.Equal(t, test.message, batch[0].Message)
+		assert.Equal(t, test.expected, batch[0].Message)
+		assert.Equal(t, test.expected, batch[0].Error.Message)
 	}
+}
+
+// If event message is non-empty, do not override with error message.
+func TestSetEventMessage(t *testing.T) {
```
carsonip, line 68: Depending on the outcome of the above discussion, the tests can be consolidated by having expectedMessage and expectedErrorMessage fields.

model/modelprocessor/errormessage.go
rubvs: Okay, then I'm miss understanding the purpose of `event.Error.Message`. Why would we not capture the error body in an error field, but potential capture it in a non-error field?

simitt: @rubvs the error message is already assigned to `event.Error.Message` in https://github.com/elastic/apm-data/blob/main/input/elasticapm/internal/modeldecoder/v2/decoder.go#L1690

1pkg (approved):

model/modelprocessor/errormessage_test.go
```diff
@@ -27,39 +27,70 @@ import (
...
 	for _, test := range tests {
-		batch := modelpb.Batch{{Error: test.input}}
+		batch := modelpb.Batch{test.input}
 		processor := modelprocessor.SetErrorMessage{}
 		err := processor.ProcessBatch(context.Background(), &batch)
 		assert.NoError(t, err)
-		assert.Equal(t, test.message, batch[0].Message)
+		assert.Equal(t, test.expectedMessage, batch[0].Message)
+		assert.Equal(t, test.expectedLogMessage, batch[0].Error.Log.Message)
+		assert.Equal(t, test.expectedExceptionMessage, batch[0].Error.Exception.Message)
```
axw, line 94: These assertions seem superfluous, since the processor doesn't modify the log/exception message.

axw (dismissed):
Thanks! I think we should probably remove the `expectedLogMessage` and `expectedExceptionMessage` bits, but otherwise LGTM.

carsonip (approved):

PR #346, opened 2024-08-20
Improve attribute handling

rubvs:
Fix https://github.com/elastic/apm-server/issues/13703, with some refactoring of the conversion between OTLP attributes and APMEvent labels.

## Context

From my understanding looking at https://opentelemetry.io/docs/specs/otel/common/#attribute, the element values of an array attribute has to:

1. Only consist of primitive values which includes; `string`, `int64`, `bool`, and `float64`
2. All elements in this array has to be the same (homogeneous).

It can be deduced that nested arrays are not allowed, since if this would be the case, the spec would have to also allow pointers, or a struct type to be included as a primitive. Effectively, creating an array _reference_. An array as _value_ would not suffice, since its not atomic in principle. Rather, a collection, therefore breaking the homogeneous invariant. **So I highly doubt that Attribute array value are allowed to be nested.** 

From a practical point of view, the concept of an attribute is not hierarchical. So nesting doesn't really makes sense. Spans and traces can be nested. Attributes cannot.

## Problem

I've come to believe the reason for the bug in https://github.com/elastic/apm-server/issues/13703 is more fundamental to the way we convert between OTLP attributes and APMEvent labels, than simply dropping heterogeneous elements.

The func `ifaceAttributeValue` acts as an interface between conversion of a `pcommon.Value` and a Go primitive. However, this is done by using the `interface` construct in Go as an abstract _buffer_ for type conversion. 

This technique, though valid, in practice tends to force a developer's hand to introduce templating, as we see in the `pSliceToType` function. Which is a slippery slope into metaprogramming hell. This is a deadly trap in C++, and is why I'm personally not a big proponent of templates in Go.

The bigger problem with introducing `interface` as a buffer, is that at some point, we have to do the actual conversion from `any` (aka `interface`) to the real, concrete Go primitives. To my limited knowledge, this is what I understand the purpose of `setLabel` to be; a producer of Go primitives when the exact translation is not made explicit. Therefore, `setLabel` is mostly called in the `default` section or the `switch` statements in translation methods, like `translateResourceMetadata`, etc.

The problem with this is multifold:

1. **Testing:** It becomes very hard to test. You're left with converting between primitives multiple times to ensure you handle all the conversion edge cases.
2. **Information Loss:** Fundamentally, the reason for this difficult testing is due to information loss by introducing `interface`. For example, the concept of a `bool` in lost after `ifaceAttributeValue` is called.
3. **Stack:** `ifaceAttributeValue` is called recursively to ensure that all array elements are properly converted. This recursive call is needed if arrays can be nested, which I don't think they can. So you are left with unnecessary stack calls.
4. **Complexity:** All of this adds unnecessary code; extra testing, `interface`, templating, more functions than needed, and difficulty debugging. 

## Solution

Do the conversion between `pcommon.Value` and Go primitives directly (by only calling `setLabel`). Do not use `interface` or templating. Ignore a recursive solution. We are not working with a Tree or Graph data structure. Simply, a linear homogeneous array.

### DIFF
--- a/input/otlp/metadata.go
+++ b/input/otlp/metadata.go
@@ -185,7 +185,6 @@ func translateResourceMetadata(resource pcommon.Resource, out *modelpb.APMEvent)
 				out.Host = modelpb.HostFromVTPool()
 			}
 			out.Host.Architecture = truncate(v.Str())
-			// TODO: Don't think this is being tested.
 		case semconv.AttributeHostIP:
 			if out.Host == nil {
 				out.Host = modelpb.HostFromVTPool()
@@ -194,10 +193,9 @@ func translateResourceMetadata(resource pcommon.Resource, out *modelpb.APMEvent)
 			result := make([]*modelpb.IP, 0, slice.Len())
 			for i := 0; i < slice.Len(); i++ {
 				ip, err := modelpb.ParseIP(slice.At(i).Str())
-				if err != nil {
-					panic(err)
+				if err == nil {
+					result = append(result, ip)
 				}
-				result = append(result, ip)
 			}
 			out.Host.Ip = result
 
@@ -503,7 +501,7 @@ func setLabel(key string, event *modelpb.APMEvent, v pcommon.Value) {
 		modelpb.NumericLabels(event.NumericLabels).Set(key, v.Double())
 	case pcommon.ValueTypeSlice:
 		s := v.Slice()
-		if v.Slice().Len() == 0 {
+		if s.Len() == 0 {
 			return
 		}
 		switch s.At(0).Type() {
--- a/input/otlp/metadata.go
+++ b/input/otlp/metadata.go
@@ -185,17 +185,21 @@ func translateResourceMetadata(resource pcommon.Resource, out *modelpb.APMEvent)
 				out.Host = modelpb.HostFromVTPool()
 			}
 			out.Host.Architecture = truncate(v.Str())
+			// TODO: Don't think this is being tested.
 		case semconv.AttributeHostIP:
 			if out.Host == nil {
 				out.Host = modelpb.HostFromVTPool()
 			}
-			out.Host.Ip = pSliceToType[*modelpb.IP](v.Slice(), func(v pcommon.Value) (*modelpb.IP, bool) {
-				ip, err := modelpb.ParseIP(v.Str())
+			slice := v.Slice()
+			result := make([]*modelpb.IP, 0, slice.Len())
+			for i := 0; i < slice.Len(); i++ {
+				ip, err := modelpb.ParseIP(slice.At(i).Str())
 				if err != nil {
-					return nil, false
+					panic(err)
 				}
-				return ip, true
-			})
+				result = append(result, ip)
+			}
+			out.Host.Ip = result
 
 		// process.*
 		case semconv.AttributeProcessPID:
@@ -320,7 +324,6 @@ func translateResourceMetadata(resource pcommon.Resource, out *modelpb.APMEvent)
 				out.DataStream = modelpb.DataStreamFromVTPool()
 			}
 			out.DataStream.Namespace = v.Str()
-
 		default:
 			if out.Labels == nil {
 				out.Labels = make(modelpb.Labels)
@@ -482,16 +485,6 @@ func cleanServiceName(name string) string {
 	return serviceNameInvalidRegexp.ReplaceAllString(truncate(name), "_")
 }
 
-func pSliceToType[T any](slice pcommon.Slice, f func(pcommon.Value) (T, bool)) []T {
-	result := make([]T, 0, slice.Len())
-	for i := 0; i < slice.Len(); i++ {
-		if v, ok := f(slice.At(i)); ok {
-			result = append(result, v)
-		}
-	}
-	return result
-}
-
 // initEventLabels initializes an event-specific labels from an event.
 func initEventLabels(e *modelpb.APMEvent) {
 	e.Labels = modelpb.Labels(e.Labels).Clone()
@@ -499,7 +492,6 @@ func initEventLabels(e *modelpb.APMEvent) {
 }
 
 func setLabel(key string, event *modelpb.APMEvent, v pcommon.Value) {
-
 	switch v.Type() {
 	case pcommon.ValueTypeStr:
 		modelpb.Labels(event.Labels).Set(key, truncate(v.Str()))
@@ -517,10 +509,8 @@ func setLabel(key string, event *modelpb.APMEvent, v pcommon.Value) {
 		switch s.At(0).Type() {
 		case pcommon.ValueTypeStr:
 			result := make([]string, 0, s.Len())
-			fmt.Println("len", s.Len())
 			for i := 0; i < s.Len(); i++ {
 				r := s.At(i)
-				fmt.Println("r:", r.Str())
 				if r.Type() == pcommon.ValueTypeStr {
 					result = append(result, r.Str())
 				}
--- a/input/otlp/metadata.go
+++ b/input/otlp/metadata.go
@@ -548,8 +548,6 @@ func setLabel(key string, event *modelpb.APMEvent, v interface{}) {
 		case float64:
 			value := make([]float64, 0, len(v))
 			for i := range v {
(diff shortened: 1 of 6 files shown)


input/otlp/traces.go
```diff
@@ -837,10 +832,11 @@ func TranslateSpan(spanKind ptrace.SpanKind, attributes pcommon.Map, event *mode
 					event.DataStream = modelpb.DataStreamFromVTPool()
 				}
 				event.DataStream.Namespace = stringval
-
 			default:
-				modelpb.Labels(event.Labels).Set(k, stringval)
+				setLabel(k, event, v)
 			}
+		default:
```
carsonip, line 838: This is not equiv to ValueTypeSlice + pcommon.ValueTypeDouble. Are you relying on `setLabel` to filter out types e.g. Map? I'd prefer we keep the previous structure, as it is slightly easier to reason about the change, given how TranslateTransaction is structured a little differently to others e.g. TranslateSpan. The change now means both `switch` contains logic, while the previous code relies on the switch in TranslateTransaction, and the `switch` within setLabel is mostly a no-op. But I don't have a strong opinion about this.

input/otlp/metadata_test.go
```diff
@@ -335,13 +337,148 @@ func TestResourceLabels(t *testing.T) {
...
+			input: func() pcommon.Map {
+				m := pcommon.NewMap()
+				s := m.PutEmptySlice("k")
+				s.FromRaw([]any{})
+				return m
+			},
+			expectedLen: 0,
+		},
+		{
+			desc: "ensure nill array does not panic",
```
carsonip, line 450:
```suggestion
			desc: "ensure nil array does not panic",
```

input/otlp/metadata_test.go
```diff
@@ -335,13 +337,148 @@ func TestResourceLabels(t *testing.T) {
...
+				NumericLabels: make(map[string]*modelpb.NumericLabelValue),
+			}
+
+			otlp.TranslateSpan(ptrace.SpanKindInternal, tC.input(), e)
+
+			labels := e.GetLabels()[key].GetValues()
+			numericals := e.GetNumericLabels()[key].GetValues()
+			length := len(labels) + len(numericals)
+
+			assert.Equal(t, tC.expectedLen, length)
```
carsonip, line 475:
```suggestion
			assert.Equal(t, tC.expectedLen, length)
```
this line is redundant given the 2 following assertions.

input/otlp/metadata.go
```diff
@@ -185,17 +185,21 @@ func translateResourceMetadata(resource pcommon.Resource, out *modelpb.APMEvent)
 				out.Host = modelpb.HostFromVTPool()
 			}
 			out.Host.Architecture = truncate(v.Str())
+			// TODO: Don't think this is being tested.
```
carsonip, line 188: Is this for "host.arch"? I think this is covered by "host" in metadata_test, but feel free to add a test if it is not covered.

input/otlp/metadata_test.go
```diff
@@ -335,13 +337,148 @@ func TestResourceLabels(t *testing.T) {
...
 	assert.Equal(t, modelpb.NumericLabels{
-		"int_array": {Global: true, Values: []float64{123, 456}},
+		"int_array": {
+			Global: true,
+			Values: []float64{123, 456},
+		},
 	}, modelpb.NumericLabels(metadata.NumericLabels))
 }
 
+func TestTranslateSpanValue(t *testing.T) {
```
carsonip, line 353: As discussed, this can be merged into `TestTranslateSpanSlice`, and we should also test TranslateTransaction.

input/otlp/metadata_test.go
```diff
@@ -335,13 +337,148 @@ func TestResourceLabels(t *testing.T) {
...
+				Event:         &modelpb.Event{},
+				Span:          &modelpb.Span{},
+				Labels:        make(map[string]*modelpb.LabelValue),
+				NumericLabels: make(map[string]*modelpb.NumericLabelValue),
+			}
+
+			otlp.TranslateSpan(ptrace.SpanKindInternal, tC.input(), e)
+
+			labels := e.GetLabels()[key].GetValues()
+			numericals := e.GetNumericLabels()[key].GetValues()
```
carsonip, line 472: We can assert the entire Labels and NumericLabels instead of just a key within it. That'll make it easier to merge with the above `TestTranslateSpanValue` test.

carsonip:
Can you also [specify in PR description](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) that this will fix https://github.com/elastic/apm-server/issues/13603 ?

carsonip (commented):
The general direction looks great. Thanks for refactoring it. The code is much cleaner and easier to understand now.

input/otlp/metadata.go
```diff
@@ -185,17 +185,21 @@ func translateResourceMetadata(resource pcommon.Resource, out *modelpb.APMEvent)
...
 			}
-			out.Host.Ip = pSliceToType[*modelpb.IP](v.Slice(), func(v pcommon.Value) (*modelpb.IP, bool) {
-				ip, err := modelpb.ParseIP(v.Str())
+			slice := v.Slice()
+			result := make([]*modelpb.IP, 0, slice.Len())
+			for i := 0; i < slice.Len(); i++ {
+				ip, err := modelpb.ParseIP(slice.At(i).Str())
 				if err != nil {
-					return nil, false
+					panic(err)
```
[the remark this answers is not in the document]
rubvs: From the original code, I don't see the usage of the error other than appending the attr item to the slice if `err == nil`. This is made more explicit and easier to read (at least for me) in the new commit.

input/otlp/metadata.go
rubvs: You're right. It wasn't obvious to me at first glance. So I made a note for me to get back too, and/or for someone to comment in the Draft, like you just did. Thank you. Also, in the future I'll use FIXME instead to be more clear.

input/otlp/traces.go
rubvs: I just noticed I didn't mirror this change in `TranslateTransaction`. My mistake. My logic here is simply that, `setLabel` populates the `APMEvent` label map when no value derivations are needed. Meaning, if no special preprocessing needs to be done to "generate" an attribute value, then we can simple set the value in APMEvent directly (after OTLP -> Go primitive conversion, which is the purpose of `setLabel`). 

This follows from my comment in this PR description: _"To my limited knowledge, this is what I understand the purpose of setLabel to be; a producer of Go primitives when the exact translation is not made explicit."_

Said differently, with my proposed change, the `switch` statement would say the following:

> If we know the value to be set in APMEvent, default to setting it via `setLabel`. However, if we have to derive the value, do so in the appropriate switch-branch.

rubvs: In general, I see a `switch` branch without a `default` as an antipattern, since your "handling" is not exhaustive. This is something Rust has done very well with their `match` equivalent.

input/otlp/metadata_test.go
rubvs: Testing for `TranslateTransaction` is a good idea. Thanks.

However, I do not see the value in merging `TestTranslateSpanValue` and `TestTranslateSpanSlice`:

1. It'll add some redundant logic to the Table Driven Test.
2. Fundamentally, a value and slice are two different data structures.
    - A value is atomic, while an array is divisible.
    - I generally like to "split" tests based around the data structure under test (if possible).
3. The `TestTranslateSpanSlice` is already lengthy and might still expand in the future if I missed edge-cases.

I'm open to follow your advice. I just want to have these conversations explicitly, so I can upgrade my thinking.

1pkg (approved):
lgtm

input/otlp/metadata_test.go
rubvs: After having a deeper look, it seems to me that `TestArrayLabels` actually already tests the underlying functions `TranslateTransaction` and `TranslateSpan`. I extended this test to ensure `setLabel` properly drops the heterogeneous elements.

Same goes for `TestResourceLabels`. I do not know how I missed this in the beginning. 🤦

PR #360, opened 2024-09-05
Fix leaking slice capacity

rubvs:
The slicing operation in `Reslice` using `slice[:n]` creates a n-length slice. However, its capacity remains the same as the initial slice. The remaining elements are still allocated in memory. The backing array of the slice still contains n-elements after the slicing operation.

As a result, the remaining space won’t be reclaimed by the GC, and we can keep a large backing array despite using only a few elements.

This change is really only practical when working with large slices. However, I'm deducing with my limit contextual understanding, that optimization is the purpose of this reslicing function.

### DIFF
--- a/input/elasticapm/internal/modeldecoder/modeldecoderutil/slice.go
+++ b/input/elasticapm/internal/modeldecoder/modeldecoderutil/slice.go
@@ -27,7 +27,9 @@ func Reslice[Slice ~[]model, model any](slice Slice, n int) Slice {
 	if n > cap(slice) {
 		slice = slices.Grow(slice, n-len(slice))
 	}
-	return slice[:n]
+	res := make(Slice, n)
+	copy(res, slice[:n])
+	return res
 }
 
 // ResliceAndPopulateNil ensures a slice of pointers has atleast


input/elasticapm/internal/modeldecoder/modeldecoderutil/slice.go
```diff
@@ -27,7 +27,9 @@ func Reslice[Slice ~[]model, model any](slice Slice, n int) Slice {
 	if n > cap(slice) {
 		slice = slices.Grow(slice, n-len(slice))
 	}
-	return slice[:n]
+	res := make(Slice, n)
+	copy(res, slice[:n])
+	return res
```
lahsivjar, line 32: This can create multiple allocations of slice (one in Grow and the other one which we are doing here). IIANM, the whole point of this function was to reduce allocation and GC (even when the capacity is higher) so I don't think this change is needed. OTOH, now that we are moving away from pooling, we might not even require Reslice but that would be a bigger change overall.

rubvs: Okay, I see. Thanks. I'll close this.

PR #369, opened 2024-09-13
Sanitize OTEL datastream

rubvs:
Fixes https://github.com/elastic/apm-server/issues/14043

@carsonip I am not exactly sure I should remove disallowed runes, or replace them with `_`.


carsonip:
> @carsonip I am not exactly sure I should remove disallowed runes, or replace them with _.

Replacing is fine, to make it consistent with ES reroute processor implementation: https://github.com/elastic/elasticsearch/blob/33d63fa9c2fc3afb9184f31e2681942cfd8dc7fd/modules/ingest-common/src/main/java/org/elasticsearch/ingest/common/RerouteProcessor.java#L203-L206

input/otlp/logs.go
```diff
@@ -239,3 +245,28 @@ func (c *Consumer) convertLogRecord(
...
+	if strings.ContainsAny(k, DisallowedDataStreamRunes) {
+		k = strings.Map(replaceReservedRune, k)
+	}
+	// Cannot start with  _, +
+	// https://github.com/elastic/ecs/blob/main/rfcs/text/0009-data_stream-fields.md
+	if k[0] == '_' || k[0] == '+' {
+		k = k[1:]
+	}
+	if utf8.RuneCountInString(k) > MaxDataStreamRunes {
+		return string([]rune(k)[:MaxDataStreamRunes])
```
carsonip, line 261: whether it should be 100 unicode chars or 100 bytes depends on discussion in https://github.com/elastic/elasticsearch/issues/112936

input/otlp/logs.go
```diff
@@ -239,3 +245,28 @@ func (c *Consumer) convertLogRecord(
...
+	}
+	return k
+}
+
+func replaceReservedRune(r rune) rune {
+	switch r {
+	case '-', '\\', '/', '*', '?', '"', '<', '>', '|', ' ', ',', '#':
```
carsonip, line 268: Having a list of disallowed runes in this switch case, as well as in DisallowDataStreamRunes creates 2 sources of truth and will be harder to maintain down the road, especially when the 2 lines are separated by 200+ lines.

input/otlp/logs.go
```diff
@@ -239,3 +245,28 @@ func (c *Consumer) convertLogRecord(
 
 	return event
 }
+
+// https://www.elastic.co/guide/en/ecs/current/ecs-data_stream.html
+func sanitizeDataStream(k string) string {
```
carsonip, line 250:
```suggestion
func sanitizeDataStreamField(k string) string {
```

nit: sanitizeDataStream(k) makes me assume that k is the full data stream name e.g. `logs-foo-default`. To make it even more explanatory, it would be `func sanitizeDataStreamField(datasetOrNamespace string) string {`, but it doesn't look idiomatic. An alternative would be to add a comment to describe what this function does, and what it expects.

carsonip, line 250: the functions should also be moved out of `logs.go`. It is not specific to logs.

input/otlp/logs.go
```diff
@@ -239,3 +245,26 @@ func (c *Consumer) convertLogRecord(
...
+
+// Sanitize the datastream fields (dataset, namespace) to apply restrictions
+// as outlined in https://www.elastic.co/guide/en/ecs/current/ecs-data_stream.html
+func sanitizeDataStreamField(field string) string {
+	field = strings.ToLower(field)
+	field = strings.Map(replaceReservedRune, field)
+	// Cannot start with  _, +
+	// https://github.com/elastic/ecs/blob/main/rfcs/text/0009-data_stream-fields.md
+	if field[0] == '_' || field[0] == '+' {
+		field = field[1:]
```
1pkg, line 257: should this check repeat until first character is not '_' or '+' ?

input/otlp/logs.go
```diff
@@ -48,6 +49,11 @@ import (
 	"github.com/elastic/apm-data/model/modelpb"
 )
 
+const (
+	MaxDataStreamRunes        = 100
```
1pkg, line 53: nit: do we want these internal naming check constants be exposed from this package?

input/otlp/logs.go
```diff
@@ -239,3 +245,26 @@ func (c *Consumer) convertLogRecord(
 
 	return event
 }
+
+// Sanitize the datastream fields (dataset, namespace) to apply restrictions
+// as outlined in https://www.elastic.co/guide/en/ecs/current/ecs-data_stream.html
+func sanitizeDataStreamField(field string) string {
```
1pkg, line 251: Also depending on how performance sensitive this function is, we could also use an ascii table map trick [link](https://go.dev/play/p/NeksqtwNJ0L). It's x7 faster on average than `strings.Map`. It's also way less readable, so I'd say it's not worth it unless this is a hot execution path.

1pkg (dismissed):

input/otlp/logs.go
```diff
@@ -239,3 +245,26 @@ func (c *Consumer) convertLogRecord(
...
+	field = strings.ToLower(field)
+	field = strings.Map(replaceReservedRune, field)
+	// Cannot start with  _, +
+	// https://github.com/elastic/ecs/blob/main/rfcs/text/0009-data_stream-fields.md
+	if field[0] == '_' || field[0] == '+' {
```
carsonip, line 256: This check is actually incorrect. The restriction is only on data stream type, as index name cannot start with `_` or `+`. There is no such restriction on dataset and namespace.

carsonip (changes requested):

input/otlp/logs.go
```diff
@@ -239,3 +245,26 @@ func (c *Consumer) convertLogRecord(
...
 	return event
 }
+
+// Sanitize the datastream fields (dataset, namespace) to apply restrictions
+// as outlined in https://www.elastic.co/guide/en/ecs/current/ecs-data_stream.html
+func sanitizeDataStreamField(field string) string {
+	field = strings.ToLower(field)
+	field = strings.Map(replaceReservedRune, field)
+	// Cannot start with  _, +
+	// https://github.com/elastic/ecs/blob/main/rfcs/text/0009-data_stream-fields.md
```
carsonip, line 255: I'd prefer a link to the actual ecs definition rather than rfc. The rfc is outdated and says `type` should either be `logs` or `metrics`.

input/otlp/logs.go
```diff
@@ -239,3 +244,21 @@ func (c *Consumer) convertLogRecord(
 
 	return event
 }
+
+// Sanitize the datastream fields (dataset, namespace) to apply restrictions
+// as outlined in https://www.elastic.co/guide/en/ecs/current/ecs-data_stream.html
+func sanitizeDataStreamField(field string) string {
```
carsonip, line 250: Can you move these util funcs out of `logs.go`?

input/otlp/metrics.go
```diff
@@ -185,7 +185,7 @@ func (c *Consumer) handleScopeMetrics(
 					if event.DataStream == nil {
```
carsonip, line 185: line 183 (2 lines above) is missing the sanitize call.

input/otlp/logs_test.go
```diff
@@ -567,6 +571,15 @@ func TestConsumerConsumeLogsDataStream(t *testing.T) {
 			expectedDataStreamDataset:   "1",
 			expectedDataStreamNamespace: "2",
 		},
+		// Test data sanitization: https://www.elastic.co/guide/en/ecs/current/ecs-data_stream.html
+		// 1. Replace all disallowed runes with _
+		// 2. Datastream length should not exceed otlp.MaxDataStreamBytes
+		{
+			resourceDataStreamDataset:   otlp.DisallowedDataStreamRunes + randomString,
+			resourceDataStreamNamespace: otlp.DisallowedDataStreamRunes + randomString,
+			expectedDataStreamDataset:   revertedString(otlp.DisallowedDataStreamRunes) + randomString[:maxLen],
```
carsonip, line 580: I don't think the test util function is needed when we can use strings.Repeat
```suggestion
			expectedDataStreamDataset:   strings.Repeat("_", len(otlp.DisallowedDataStreamRunes)) + randomString[:maxLen],
```

input/otlp/logs_test.go
```diff
@@ -532,6 +532,10 @@ func processLogEvents(t *testing.T, logs plog.Logs) modelpb.Batch {
 }
 
 func TestConsumerConsumeLogsDataStream(t *testing.T) {
+	randomString, err := generateRandomString(otlp.MaxDataStreamBytes)
```
carsonip, line 535: `generateRandomString` can be replaced by strings.Repeat("abcdefghijklmnopqrstuvwxyz0123456789", 10) then we can get rid of the test util func, remove crypto/rand import, as well as error handling from generating a random string

input/otlp/metadata.go
```diff
@@ -545,3 +547,21 @@ func setLabel(key string, event *modelpb.APMEvent, v pcommon.Value) {
...
+		return field[:MaxDataStreamBytes]
+	}
+	return field
+}
+
+func replaceReservedRune(r rune) rune {
+	if strings.ContainsRune(DisallowedDataStreamRunes, r) {
+		return '_'
+	}
+	return r
```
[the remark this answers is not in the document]
carsonip: personally i'd switch case a rune (or byte behind a multi-byte rune check), hoping that compiler is smart enough to create a jump table. Otherwise, @1pkg 's ascii table map trick is as good.

> Having a list of disallowed runes in this switch case, as well as in DisallowDataStreamRunes creates 2 sources of truth and will be harder to maintain down the road, especially when the 2 lines are separated by 200+ lines.

Revisiting my earlier comment, @rubvs actually you have a solution very close to what I'd do initially. My duplication concern comes from 2 sources of truth in actual code file. But if the 2 sources are spread across actual code and test code, I'd be less concerned.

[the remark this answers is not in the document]
lahsivjar: > personally i'd switch case a rune (or byte behind a multi-byte rune check), hoping that compiler is smart enough to create a jump table

I think the issue here is optimizing the comparison. A rangetable would create efficient comparison and should perform quite well if the number of runes to compare is high enough, however, since we have only a few here the overheads would probably drown any gains or make it comparable. We are probably talking about a few nanosecs here so not much of a difference anyway. Not sure how the switch-case will look like though since it would technically be all the runes in one case.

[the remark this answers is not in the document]
carsonip: > Not sure how the switch-case will look like though since it would technically be all the runes in one case.

That's how I imagine.

+1 on performance difference here. I'm fine with the ContainsRune here since my bench also shows ns-level baseline

carsonip (approved):

carsonip (commented):
After discussing with @felixbarny , we agreed that it will be easier to push for a change to loosen up the ecs spec such that namespace can contain `-`, which is how es reroute processor is implemented, rather than to push a breaking change in es. Therefore, can you change the implementation in this PR to allow `-` in namespace?

rubvs:
> After discussing with @felixbarny , we agreed that it will be easier to push for a change to loosen up the ecs spec such that namespace can contain `-`, which is how es reroute processor is implemented, rather than to push a breaking change in es. Therefore, can you change the implementation in this PR to allow `-` in namespace?

Just to make sure, the `-` is still disallowed in `dataset`?

carsonip:
> Just to make sure, the - is still disallowed in dataset?

Correct. Otherwise parsing dataset and namespace would be ambiguous

1pkg (dismissed):

carsonip (approved):
lgtm, thanks for bearing with me!

Issue #361, opened 2024-09-05
Flaky semaphore test


rubvs:
Test failure when running `make test` with `-race` enabled:

```
=== RUN   TestConsumeLogsSemaphore
panic: test timed out after 10m0s
running tests:
	TestConsumeLogsSemaphore (10m0s)

goroutine 130 [running]:
testing.(*M).startAlarm.func1()
	/usr/local/go/src/testing/testing.go:2366 +0x265
created by time.goFunc
	/usr/local/go/src/time/sleep.go:177 +0x45

goroutine 1 [chan receive, 9 minutes]:
testing.(*T).Run(0xc0003d41a0, {0x103362ede, 0x18}, 0x1037051e0)
	/usr/local/go/src/testing/testing.go:1750 +0x851
testing.runTests.func1(0xc0003d41a0)
	/usr/local/go/src/testing/testing.go:2161 +0x86
testing.tRunner(0xc0003d41a0, 0xc0004dfb10)
	/usr/local/go/src/testing/testing.go:1689 +0x21f
testing.runTests(0xc0003569d8, {0x103daec40, 0x46, 0x46}, {0xc0004dfbb8?, 0xc0004dfc00?, 0x103e076e0?})
	/usr/local/go/src/testing/testing.go:2159 +0x8bf
testing.(*M).Run(0xc00036d2c0)
	/usr/local/go/src/testing/testing.go:2027 +0xf18
main.main()
	_testmain.go:187 +0x2be

goroutine 123 [chan receive, 9 minutes]:
github.com/elastic/apm-data/input/otlp_test.TestConsumeLogsSemaphore.func1({0xc0000d1c48?, 0x102416989?}, 0xc0005657a0)
	/Users/rubenvanstaden/github/elastic/apm-data/input/otlp/logs_test.go:231 +0x51
github.com/elastic/apm-data/model/modelpb.ProcessBatchFunc.ProcessBatch(0xc0001209f0, {0x103716b58, 0xc000132bd0}, 0xc0005657a0)
	/Users/rubenvanstaden/github/elastic/apm-data/model/modelpb/batch.go:42 +0x48
github.com/elastic/apm-data/input/otlp.(*Consumer).ConsumeLogsWithResult(0xc000161ec0, {0x103716b58, 0xc000132bd0}, {0xc0001209c0?, 0xc000664d2c?})
	/Users/rubenvanstaden/github/elastic/apm-data/input/otlp/logs.go:79 +0x5ca
[...]

PR #391, opened 2024-11-04
Fix Flaky Semaphore Test

rubvs:
Closes https://github.com/elastic/apm-data/issues/361

## Context

Testing `apm-data` with `-race` causes a deadlock, since the semaphore is not properly released.

The [TestConsumeLogsSemaphore](https://github.com/elastic/apm-data/blob/46f8e38486b772090fbf1fdccc8eeea13c513075/input/otlp/logs_test.go#L226) basically consists of 3 parts:

1. Spin up a goroutine that runs a consumer in parallel with the main routine. The purpose of this child routine is to acquire the semaphore until [close(doneCh)](https://github.com/elastic/apm-data/blob/46f8e38486b772090fbf1fdccc8eeea13c513075/input/otlp/logs_test.go#L254).
2. The main routine spins up its own consumer, which after attempting for some time to acquire the semaphore, will return with an expected error.
3. After the expected error, finish running the parallel routine in (1) to release the semaphore. Main runs a last consumer to ensure it can properly acquire the released lock.

## Solution

Removing the following lines from function [semAcquire](https://github.com/elastic/apm-data/blob/46f8e38486b772090fbf1fdccc8eeea13c513075/input/otlp/trace_semaphore.go#L29) seems to fix the issue.
```
sp, ctx := apm.StartSpan(ctx, "Semaphore.Acquire", "Reporter")
defer sp.End()
```

Initially, I thought the problem was _propagating an inappropriate context_, but passing a background context and removing the context shadowing, didn't solve the problem. 

The only reasonable explanation, therefore, is the span introduces enough of a lag such that [Goroutine 1](https://github.com/elastic/apm-data/blob/46f8e38486b772090fbf1fdccc8eeea13c513075/input/otlp/logs_test.go#L245) does not acquire the semaphore first, before [Main](https://github.com/elastic/apm-data/blob/46f8e38486b772090fbf1fdccc8eeea13c513075/input/otlp/logs_test.go#L252) starts competing for it, leading to a  deadlock.

The deadlock is solved by "upgrading" the [startCh](https://github.com/elastic/apm-data/blob/46f8e38486b772090fbf1fdccc8eeea13c513075/input/otlp/logs_test.go#L242) to sync the parallel running goroutine with main, not when it starts, but when it has fully acquired the semaphore. This is the purpose of `semAcquiredCh`.

However, since the processor is executed multiple times, for the context of this test we only want to sync the first acquisition. Hence, the use of `sync.Once`. 

## Update

The tests were update to remove `sync.Once` using the suggestion by @1pkg in the comments below: https://github.com/elastic/apm-data/pull/391#discussion_r1830180915

## Test

```
go clean -testcache
go test -race -v -failfast ./input/otlp
```

### DIFF
--- a/input/otlp/logs_test.go
+++ b/input/otlp/logs_test.go
@@ -226,7 +226,6 @@ func TestConsumerConsumeLogs(t *testing.T) {
 
 func TestConsumeLogsSemaphore(t *testing.T) {
 	logs := plog.NewLogs()
-	var batches []*modelpb.Batch
 
 	var once sync.Once
 	semAcquiredCh := make(chan struct{})
@@ -238,8 +237,6 @@ func TestConsumeLogsSemaphore(t *testing.T) {
 			close(semAcquiredCh)
 		})
 		<-doneCh
-		batchCopy := batch.Clone()
-		batches = append(batches, &batchCopy)
 		return nil
 	})
 	consumer := otlp.NewConsumer(otlp.ConsumerConfig{
--- a/input/otlp/logs_test.go
+++ b/input/otlp/logs_test.go
@@ -38,7 +38,6 @@ import (
 	"context"
 	"fmt"
 	"strings"
-	"sync"
 	"testing"
 	"time"
 
@@ -227,16 +226,11 @@ func TestConsumerConsumeLogs(t *testing.T) {
 func TestConsumeLogsSemaphore(t *testing.T) {
 	logs := plog.NewLogs()
 
-	var once sync.Once
-	semAcquiredCh := make(chan struct{})
 	doneCh := make(chan struct{})
-
 	recorder := modelpb.ProcessBatchFunc(func(ctx context.Context, batch *modelpb.Batch) error {
 		// Ensure channel is only closed the first time
-		once.Do(func() {
-			close(semAcquiredCh)
-		})
-		<-doneCh
+		doneCh <- struct{}{}
+		doneCh <- struct{}{}
 		return nil
 	})
 	consumer := otlp.NewConsumer(otlp.ConsumerConfig{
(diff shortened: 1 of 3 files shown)


input/otlp/logs_test.go
```diff
@@ -225,34 +226,43 @@ func TestConsumerConsumeLogs(t *testing.T) {
 
 func TestConsumeLogsSemaphore(t *testing.T) {
 	logs := plog.NewLogs()
-	var batches []*modelpb.Batch
 
+	var once sync.Once
```
1pkg, line 230: I think, we don't really need this complicated concurrency primitive structure here and can simplify this. The same test can be accomplished with a single buffered channel without any additional primitives, see the example below.

```
func TestConsumeLogsSemaphore(t *testing.T) {
	logs := plog.NewLogs()
	var batches []*modelpb.Batch

	doneCh := make(chan struct{})
	recorder := modelpb.ProcessBatchFunc(func(ctx context.Context, batch *modelpb.Batch) error {
		doneCh <- struct{}{}
		doneCh <- struct{}{}
		batchCopy := batch.Clone()
		batches = append(batches, &batchCopy)
		return nil
	})
	consumer := otlp.NewConsumer(otlp.ConsumerConfig{
		Processor: recorder,
		Semaphore: semaphore.NewWeighted(1),
	})

	go func() {
		_, err := consumer.ConsumeLogsWithResult(context.Background(), logs)
		assert.NoError(t, err)
	}()

	<-doneCh
	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond)
	defer cancel()
	_, err := consumer.ConsumeLogsWithResult(ctx, logs)
	assert.Equal(t, err.Error(), "context deadline exceeded")
	<-doneCh

	// turn channel into sink
	doneCh = make(chan struct{}, 2)
	_, err = consumer.ConsumeLogsWithResult(context.Background(), logs)
	assert.NoError(t, err)
}
```

1pkg (approved):

PR #395, opened 2024-11-07
fix: use go.mod protobuf version for generation

kruskall:
## 🧑‍💻 What changed

remove explicit version from shell script and use go.mod

## ❓ Why was this changed

go.mod version and shell version needs to be kept in sync

## ✅ How to test

run `make generate-modelpb`


rubvs (approved):

model/modelpb/useragent_vtproto.pb.go
```diff
@@ -16,7 +16,7 @@
 // under the License.
 
 // Code generated by protoc-gen-go-vtproto. DO NOT EDIT.
-// protoc-gen-go-vtproto version: v0.6.0
+// protoc-gen-go-vtproto version: v0.6.1-0.20240319094008-0393e58bdf10
```
simitt, line 19: Is this now using an unreleased version ?

PR #409, opened 2024-12-06
Expand Result to Include Signal Types

rubvs:
Related to solving https://github.com/elastic/apm-managed-service/issues/606 for incoming APM events (not needed for OTEL).

## Context

We need to encode the incoming signal type into the `Result` datastructure, which we can then use within ingest service and index service to extend the necessary metric attributes.

Fundamentally, [Result](https://github.com/elastic/apm-data/blob/603774651d9ff2329fc4c0966db744092b70d25a/input/elasticapm/result.go#L24) is an abstraction that encodes information after we have processed a set of events, called a _batch_. It operates on the _level_ of a batch not an event.

Therefore, we don't have information on the event level. We need to update `Result` to encode signal types. There are 3 ways of doing this: 

1. Update the `Accepted`, and related, fields to a map datastructure
2. Create a `Signal` struct, or something similar, to containerize data
3. Manually add the different fields

## Solution

Updating the current fields (1) will introduce breaking changes to both APM Server and MIS, since `Result` operates on the protocol level. Creating a new struct (2) is redundant, but I"m open to suggestions. Therefore, manually adding fields (3) is the simplest change.


input/elasticapm/processor.go
```diff
@@ -296,19 +296,25 @@ func (p *Processor) handleStream(
...
+	}
 	result.Accepted += n
 	return readErr
 }
 
-// processBatch processes the batch and returns the events to the pool after it's been processed.
-func (p *Processor) processBatch(ctx context.Context, processor modelpb.BatchProcessor, batch *modelpb.Batch) error {
-	return processor.ProcessBatch(ctx, batch)
-}
-
```
rubvs, line 311: **Reviewers**: I do not see any use for this method.

marclop: I think it's a leftover from when we introduced async processing, but we never removed it.

input/elasticapm/processor.go
```diff
@@ -296,19 +296,25 @@ func (p *Processor) handleStream(
 	if n == 0 {
 		return readErr
 	}
-
-	if err := p.processBatch(ctx, processor, batch); err != nil {
+	if err := processor.ProcessBatch(ctx, batch); err != nil {
 		return fmt.Errorf("cannot process batch: %w", err)
 	}
+	for _, v := range *batch {
```
marclop, line 302: Can you add some unit tests where these counters are verified?

input/elasticapm/result.go
```diff
@@ -23,18 +23,34 @@ import (
...
 	// counters above are still incremented.
 	Errors []error
+
 	// Accepted holds the number of valid events accepted.
-	Accepted int
+	Accepted            int
+	SpanAccepted        int
+	TransactionAccepted int
+	MetricAccepted      int
+	LogAccepted         int
```
marclop, line 37: I have a suggestion to make all these counters a bit more digestible:

```go
type Result struct {
	errorsSpace [5]error

	// Errors holds a limited number of errors that occurred while
	// processing the event stream. If the limit is reached, the
	// counters above are still incremented.
	Errors []error

	// Accepted holds the number of valid events accepted.
	Accepted        int
	AcceptedDetails ProcessedDetail

	// TooLarge holds the number of events that were rejected due
	// to exceeding the event size limit.
	TooLarge        int
	TooLargeDetails ProcessedDetail

	// Invalid holds the number of events that were rejected due
	// to being invalid, excluding those that are counted by TooLarge.
	Invalid        int
	InvalidDetails ProcessedDetail
}

// ProcessedDetail holds the number of events processed for each type.
type ProcessedDetail struct {
	Transaction int
	Span        int
	Metric      int
	Log         int
	Error       int
}
```

However, I don't see where the TooLarge / Invalid counters are incremented, did you forget to include the patch?

input/elasticapm/processor.go
```diff
@@ -296,19 +296,25 @@ func (p *Processor) handleStream(
...
+		switch v.Type() {
+		case modelpb.SpanEventType:
+			result.SpanAccepted++
+		case modelpb.TransactionEventType:
+			result.TransactionAccepted++
+		case modelpb.MetricEventType:
+			result.MetricAccepted++
+		case modelpb.LogEventType:
```
marclop, line 310: `modelpb.ErrorEventType` seem to be missing.

input/elasticapm/result.go
1pkg: Agree with Mark here, also since it's public type, for all export fields we should add a godoc comment // with a short description.

rubvs: @marclop I'll use the `ProcessedDetail` approach. I was actually thinking of not having a detailed breakdown for  `TooLarge` and `Invalid`, since both error types are basically saying there is an error in the body or in reading the body properly. So encoding the signal type makes no sense, since there might not a signal. For example, from [L214](https://github.com/elastic/apm-data/blob/485554733eceacce688dec287801b1e699c2c7bf/input/elasticapm/processor.go#L214), an Invalid error is captured if no signal type could be deduced.

rubvs: Relating to the issue this PR is solving: We only want to have a detailed breakdown of the _accepted_ events in our dashboards.

input/elasticapm/processor.go
rubvs: Extended test `TestHandleStream` to check proper population.

input/elasticapm/result.go
```diff
@@ -23,20 +23,34 @@ import (
...
+
 	// Errors holds a limited number of errors that occurred while
 	// processing the event stream. If the limit is reached, the
 	// counters above are still incremented.
 	Errors []error
+
 	// Accepted holds the number of valid events accepted.
-	Accepted int
+	Accepted        int
+	AcceptedDetails ProcessedDetail
```
1pkg, line 34: We still should add a godoc comment for new `AcceptedDetails` field here

1pkg (approved):

input/elasticapm/result.go
rubvs: Done! Thanks

input/elasticapm/result.go
```diff
@@ -23,20 +23,37 @@ import (
...
 	// to exceeding the event size limit.
 	TooLarge int
+
 	// Invalid holds the number of events that were rejected due
 	// to being invalid, excluding those that are counted by TooLarge.
 	Invalid int
 }
 
+// ProcessedDetail holds the number of events processed for each type.
+type ProcessedDetail struct {
```
marclop, line 49: nit: should this be called `AcceptedDetail`?

marclop (approved):

marclop (approved):
Thanks!

1pkg (approved):

PR #414, opened 2024-12-17
Add specific type and subtype for otlp gen_ai spans

xrmx:
This matches what has been done in https://github.com/elastic/opentelemetry-lib/pull/127/

### DIFF
--- a/input/otlp/traces.go
+++ b/input/otlp/traces.go
@@ -615,6 +615,10 @@ func TranslateSpan(spanKind ptrace.SpanKind, attributes pcommon.Map, event *mode
 		rpcService string
 	)
 
+	var (
+		genAiSystem string
+	)
+
 	var http modelpb.HTTP
 	var httpRequest modelpb.HTTPRequest
 	var httpResponse modelpb.HTTPResponse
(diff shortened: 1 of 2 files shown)


input/otlp/traces.go
```diff
@@ -812,6 +816,11 @@ func TranslateSpan(spanKind ptrace.SpanKind, attributes pcommon.Map, event *mode
 				// stacktrace is expected to be large thus un-truncated value is needed
 				event.Code.Stacktrace = v.Str()
 
+			// gen_ai.*
+			case "gen_ai.system":
```
endorama, line 820: nitpick: I would suggest to move this to a constant.

Here the change would be:
```suggestion
			case attributeGenAiSystem:
```

and the constant should be added to the beginning of the file, below https://github.com/xrmx/apm-data/blob/64e886dbca1d5ae08ac23b6e7862942c4fc65859/input/otlp/traces.go#L66-L84

endorama (approved):

input/otlp/traces.go
xrmx: oh, missed that, updated!

endorama (approved):

endorama:
Thank you!

PR #416, opened 2024-12-18
Translate OTEL log's InstrumentationScope as logger name

oldium:
InstrumentationScope usually contains the name of the instrumentation library. However, in the case of logs, it may also contain the name of the logger if the log source defines it, see [here][1]. An instrumentation library usually has a version set, so assume that the name without the version is the name of the logger.

[1]: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/api.md#get-a-logger

### DIFF
--- a/input/otlp/logs.go
+++ b/input/otlp/logs.go
@@ -119,7 +119,7 @@ func (c *Consumer) convertLogRecord(
 	event := baseEvent.CloneVT()
 	initEventLabels(event)
 
-	translateScopeMetadata(scope, event)
+	translateLogScopeMetadata(scope, event)
 
 	if record.Timestamp() == 0 {
 		event.Timestamp = modelpb.FromTime(record.ObservedTimestamp().AsTime().Add(timeDelta))
--- a/input/otlp/metadata.go
+++ b/input/otlp/metadata.go
@@ -456,7 +456,7 @@ func translateResourceMetadata(resource pcommon.Resource, out *modelpb.APMEvent)
 	}
 }
 
-func translateScopeMetadata(scope pcommon.InstrumentationScope, out *modelpb.APMEvent) {
+func translateScopeCommonMetadata(scope pcommon.InstrumentationScope, out *modelpb.APMEvent) {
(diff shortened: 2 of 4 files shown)


oldium:
Signed the agreement

oldium:
Anybody here?

carsonip (commented):
Sorry for the delayed response. I'll have to check with the team and see whether introducing an assumption around "a missing version = scope name is a logger name" is desirable. Meanwhile, do you mind adding a unit test for the logic?

oldium:
> I'll have to check with the team and see whether introducing an assumption around "a missing version = scope name is a logger name" is desirable.

Yes, please. The Instrumentation Scope field is unfortunately shared with Logger Name as [intended](https://github.com/open-telemetry/opentelemetry-specification/pull/2359)...

> Meanwhile, do you mind adding a unit test for the logic?

Yes, sure.

carsonip:
@oldium thanks for the PR. While this PR fixes your use case, it might be a breaking change for some users relying on this behavior. In fact, this is an inherent issue stemming from how apm server converts some known OTel fields to the closest APM field.

At Elastic, we are actively working on an OTel-native ingestion solution named EDOT as an alternative to APM server. See [docs for EDOT](https://github.com/elastic/opentelemetry) and [a blog post for EDOT collector](https://www.elastic.co/observability-labs/blog/elastic-distribution-opentelemetry-collector) for more details. EDOT collector accepts otlp, sends to Elasticsearch directly, bypassing APM Server, and stores them in an OpenTelemetry native format, e.g. scope name will be stored in `scope.name`, resource attributes will be under `resource.attributes.*`, while having APM UI work out of the box.

Alternatively, if you prefer to stick with APM Server, you may set an ingest pipeline for logs to unset `service.framework.name` and set `log.logger` if `service.framework.version == nil`, which would yield the same result as your PR. See [docs on ingest pipelines](https://www.elastic.co/guide/en/observability/8.17/apm-ingest-pipelines.html).

Closing this PR but let me know if you have any questions.

Issue #146, opened 2020-01-22
Investigate switching error grouping key calculation to use something faster


axw:
We should investigate switching from MD5 to something faster, like [xxhash](https://cyan4973.github.io/xxHash/), for calculating error grouping keys. MD5 is cryptographic, which is not necessary for our purposes. xxHash is non-cryptographic, and considerably faster while maintaining high quality hashes.

PR #423, opened 2025-01-09
Make SetGroupingKey's hash configurable

axw:
Allow callers to specify the hash algorithm. Part of https://github.com/elastic/apm-data/issues/146

NOTE: this is a breaking change to the API: callers will need to be updated to specify the NewHash field or they will panic.

### DIFF
--- a/model/modelprocessor/groupingkey.go
+++ b/model/modelprocessor/groupingkey.go
@@ -19,7 +19,6 @@ package modelprocessor
 
 import (
 	"context"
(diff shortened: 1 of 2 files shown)


1pkg (approved):
Should be merged with caution, since it requires major version release due to breaking change.

axw:
Holding off on merging until we have updated services using this code.

carsonip (approved):

PR #431, opened 2025-01-24
feat: Add user fields mapping to OTLP

ericywl:
## ❓ Why is this being changed
Related to elastic/apm-server#15254

## 🧑‍💻 What is being changed
Map the following user fields from OTel attributes to `APMEvent.User`
- `user.id`
- `user.name`
- `user.email`

The other user fields do not have equivalents in `APMEvent.User`, so they will still be under `APMEvent.Labels`.

## ✅ How to validate the change
Run test on APM server with the changes

### DIFF
--- a/input/otlp/metadata.go
+++ b/input/otlp/metadata.go
@@ -242,8 +242,6 @@ func translateResourceMetadata(resource pcommon.Resource, out *modelpb.APMEvent)
 			}
 			out.User.Name = truncate(v.Str())
 
-		// TODO(eric): Check if `user.*` fields are considered resource metadata?
-
 		// os.*
 		case semconv.AttributeOSType:
 			if out.Host == nil {
@@ -474,8 +472,6 @@ func translateScopeMetadata(scope pcommon.InstrumentationScope, out *modelpb.APM
 			out.DataStream.Namespace = sanitizeDataStreamNamespace(v.Str())
 		}
 
-		// TODO(eric): Check if `user.*` fields are considered scope metadata?
-
 		return true
 	})
 
@@ -588,16 +584,16 @@ func replaceReservedRune(disallowedRunes string) func(r rune) rune {
 
 // NOTE: This will overwrite user fields set by `user.name=process.owner` from `otlp/metadata.go:translateResourceMetadata`.
 func addUserFields(attrKey string, attrVal pcommon.Value, event *modelpb.APMEvent) {
-	if updateEvent.User == nil {
-		updateEvent.User = &modelpb.User{}
+	if event.User == nil {
+		event.User = &modelpb.User{}
 	}
 
 	switch attrKey {
 	case attributeUserID:
-		updateEvent.User.Id = truncate(attrVal.Str())
+		event.User.Id = truncate(attrVal.Str())
 	case attributeUserName:
-		updateEvent.User.Name = truncate(attrVal.Str())
+		event.User.Name = truncate(attrVal.Str())
 	case attributeUserEmail:
-		updateEvent.User.Email = truncate(attrVal.Str())
+		event.User.Email = truncate(attrVal.Str())
 	}
 }
--- a/input/otlp/metadata.go
+++ b/input/otlp/metadata.go
@@ -242,6 +242,8 @@ func translateResourceMetadata(resource pcommon.Resource, out *modelpb.APMEvent)
 			}
 			out.User.Name = truncate(v.Str())
 
+		// TODO(eric): Check if `user.*` fields are considered resource metadata?
+
 		// os.*
 		case semconv.AttributeOSType:
 			if out.Host == nil {
@@ -471,6 +473,9 @@ func translateScopeMetadata(scope pcommon.InstrumentationScope, out *modelpb.APM
 			}
 			out.DataStream.Namespace = sanitizeDataStreamNamespace(v.Str())
 		}
+
+		// TODO(eric): Check if `user.*` fields are considered scope metadata?
+
 		return true
 	})
 
--- a/input/otlp/metadata.go
+++ b/input/otlp/metadata.go
@@ -580,3 +580,19 @@ func replaceReservedRune(disallowedRunes string) func(r rune) rune {
 		return unicode.ToLower(r)
 	}
 }
+
+// NOTE: This will overwrite user fields set by `user.name=process.owner` from `otlp/metadata.go:translateResourceMetadata`.
+func addUserFields(attrKey string, attrVal pcommon.Value, updateEvent *modelpb.APMEvent) {
+	if updateEvent.User == nil {
+		updateEvent.User = &modelpb.User{}
+	}
+
+	switch attrKey {
+	case attributeUserID:
+		updateEvent.User.Id = truncate(attrVal.Str())
+	case attributeUserName:
+		updateEvent.User.Name = truncate(attrVal.Str())
+	case attributeUserEmail:
+		updateEvent.User.Email = truncate(attrVal.Str())
+	}
+}
--- a/input/otlp/metadata.go
+++ b/input/otlp/metadata.go
@@ -587,7 +587,7 @@ func replaceReservedRune(disallowedRunes string) func(r rune) rune {
 }
 
 // NOTE: This will overwrite user fields set by `user.name=process.owner` from `otlp/metadata.go:translateResourceMetadata`.
-func addUserFields(attrKey string, attrVal pcommon.Value, updateEvent *modelpb.APMEvent) {
+func addUserFields(attrKey string, attrVal pcommon.Value, event *modelpb.APMEvent) {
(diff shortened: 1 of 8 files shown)


input/otlp/metadata.go
```diff
@@ -580,3 +580,19 @@ func replaceReservedRune(disallowedRunes string) func(r rune) rune {
 		return unicode.ToLower(r)
 	}
 }
+
+// NOTE: This will overwrite user fields set by `user.name=process.owner` from `otlp/metadata.go:translateResourceMetadata`.
```
ericywl, line 584: This requires some consideration, how should we handle it if the `user.name` was already set by `process.owner`? 

ref: https://github.com/elastic/apm-data/blob/99f84c00b79b6213eb5bceffc6b77630da2f629e/input/otlp/metadata.go#L239

input/otlp/metadata_test.go
```diff
@@ -413,3 +414,75 @@ func transformResourceMetadata(t *testing.T, resourceAttrs map[string]interface{
...
+				userName:  "hello",
+				userEmail: "hello@example.com",
+			},
+			expected: userField{
+				userID:    "123",
+				userName:  "hello",
+				userEmail: "hello@example.com",
+			},
+		},
+		{
```
ericywl, line 446: Here are the test cases for when the `process.owner` resource metadata attribute is set.

input/otlp/metadata.go
ericywl: See test cases at https://github.com/elastic/apm-data/pull/431/files#r1928194556

input/otlp/metadata_test.go
```diff
@@ -413,3 +414,75 @@ func transformResourceMetadata(t *testing.T, resourceAttrs map[string]interface{
...
+				userID:    "123",
+				userName:  "hello",
+				userEmail: "hello@example.com",
+			},
+			expected: userField{
+				userID:    "123",
+				userName:  "hello",
+				userEmail: "hello@example.com",
+			},
+		},
```
ericywl, line 473: Here are the test cases for when the process.owner resource metadata attribute is set.

input/otlp/metadata.go
```diff
@@ -242,6 +242,8 @@ func translateResourceMetadata(resource pcommon.Resource, out *modelpb.APMEvent)
 			}
 			out.User.Name = truncate(v.Str())
 
+		// TODO(eric): Check if `user.*` fields are considered resource metadata?
```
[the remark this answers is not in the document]
carsonip: I see that you are currently only doing it for log records, data points and span **events**. imo it is missing for spans (see TranslateTransaction and TranslateSpan) which is most likely where these attributes are.

And you wonder if it should apply to resource and scope. I think it is unlikely for scope, but not too sure about resource. @axw may have a better idea here.

input/otlp/metadata_test.go
carsonip: this is the case for record-level (log records, data points, spans) attributes overwriting resource level attributes, which is fine. But you _may_ also need to test resource level User semconv fields competing with `process.owner`. However, I'm not sure if we should expect it at the resource level. See the above comment.

input/otlp/metadata.go
```diff
@@ -580,3 +585,19 @@ func replaceReservedRune(disallowedRunes string) func(r rune) rune {
 		return unicode.ToLower(r)
 	}
 }
+
+// NOTE: This will overwrite user fields set by `user.name=process.owner` from `otlp/metadata.go:translateResourceMetadata`.
+func addUserFields(attrKey string, attrVal pcommon.Value, updateEvent *modelpb.APMEvent) {
```
carsonip, line 590:
```suggestion
func addUserFields(attrKey string, attrVal pcommon.Value, event *modelpb.APMEvent) {
```

nit: to be consistent with the rest of the code, `event` or `out` seems to be a better variable name

input/otlp/traces_test.go
```diff
@@ -1401,6 +1401,45 @@ func TestSpanCodeStacktrace(t *testing.T) {
...
+			}
+			if tc.input.userName != "" {
+				spanEvent.Attributes().PutStr("user.name", tc.input.userName)
+			}
+			if tc.input.userEmail != "" {
+				spanEvent.Attributes().PutStr("user.email", tc.input.userEmail)
+			}
+
+			allEvents := transformTraces(t, traces)
+			events := (*allEvents)[1:]
```
carsonip, line 1431: Why is the first event ignored? Is that the span itself?

carsonip (commented):
On the point about user fields being experimental, imo it is fine to handle them and map them to the most appropriate ECS fields, as we have been doing that in apm-server. The implication of this PR is that User semconv fields will be mapped differently, which makes it possibly a "breaking change".

For apm-server, if we're only doing that in a major version change, it is fine, but it also means an apm-data major release and any other apm-data changes to apm-server 8.x will require backport to the previous apm-data major.

For MIS, there's likely a different policy for breaking changes.

Therefore, I'd appreciate @simitt's views on whether this should really be considered a breaking change. It seems like a slippery slope to me.

input/otlp/metadata.go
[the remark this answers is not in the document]
ericywl: Thanks. From what I can tell, `TranslateTransaction` and `TranslateSpan` is also used by intake v2. Will this change overwrite the user fields that are set [here](https://github.com/elastic/apm-data/blob/58bbb4af96c5682ae5a03f131200265057fc3779/input/elasticapm/internal/modeldecoder/v2/decoder.go#L1746-L1766)?

input/otlp/metadata.go
carsonip: Raised in https://github.com/elastic/apm-server/issues/15254#issuecomment-2615437536 that `process.owner` mapping to `user.name` seems a bit out of place. I wonder it should actually be mapped to `process.user.name` instead.

input/otlp/metadata.go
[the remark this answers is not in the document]
axw: > I see that you are currently only doing it for log records, data points and span events. imo it is missing for spans (see TranslateTransaction and TranslateSpan) which is most likely where these attributes are.

++

> And you wonder if it should apply to resource and scope. I think it is unlikely for scope, but not too sure about resource. @axw may have a better idea here.

I'm not aware of any cases where `user` would exist at the resource level. I guess there could be some kind of resource owner attributes, but I'm only aware of the need to represent users of the instrumented service, which would be at the operation level.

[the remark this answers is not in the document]
ericywl: Got it, I will just add the change to `TranslateTransaction` and `TranslateSpan`.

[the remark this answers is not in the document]
ericywl: > Thanks. From what I can tell, `TranslateTransaction` and `TranslateSpan` is also used by intake v2. Will this change overwrite the user fields that are set [here](https://github.com/elastic/apm-data/blob/58bbb4af96c5682ae5a03f131200265057fc3779/input/elasticapm/internal/modeldecoder/v2/decoder.go#L1746-L1766)?

@axw @carsonip Will this part be a concern?

input/otlp/traces_test.go
ericywl: Yes I believe so.

carsonip: It shouldn't be ignored. We should ensure that the span document has user fields as well.

input/otlp/metadata.go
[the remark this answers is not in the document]
carsonip: As discussed in https://github.com/elastic/apm-data/pull/431#discussion_r1930363944 , I think we should really fix `user.name=process.owner` in either this PR or a separate PR as a blocker. `process.owner` shouldn't be setting `user.name` in the first place. When that's fixed, this PR won't be overwriting user fields.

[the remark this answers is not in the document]
simitt: +1 on this suggestion. @ericywl can you put up a PR to changing the current `user` mapping as a bugfix and then get back to this PR?

simitt:
>For MIS, there's likely a different policy for breaking changes.

>Therefore, I'd appreciate @simitt's views on whether this should really be considered a breaking change. It seems like a slippery slope to me.

Apologies for the late response. Correcting the current mapping of the `user.*` attributes is a bugfix which we should get in. 
For the new mapping of `user.*` fields, we need alignment with @mlunadia on how to treat otel mappings in APM Server and MIS. Let's put this on hold for some more days until we have figured this out.

simitt:
Let's close this PR, as we have closed the underlying issue as _Won't fix_. As raised by @lahsivjar , the currently mapped `user` fields are used in the Observability UIs with the current semantics of `process owner`. With the focus on otel native, let's not break this behaviour, but rather ensure the mapping is proper for the otel native solution.

PR #434, opened 2025-01-27
fix: Truncate string slice attributes in OTLP labels

ericywl:
## ❓ Why is this being changed
Related to elastic/apm-server#14716

## 🧑‍💻 What is being changed
- Truncate the strings in slices when setting label attributes

## ✅ How to validate the change
Run test on APM server with the changes

### DIFF
--- a/input/otlp/metadata.go
+++ b/input/otlp/metadata.go
@@ -516,7 +516,7 @@ func setLabel(key string, event *modelpb.APMEvent, v pcommon.Value) {
 			for i := 0; i < s.Len(); i++ {
 				r := s.At(i)
 				if r.Type() == pcommon.ValueTypeStr {
-					result = append(result, r.Str())
+					result = append(result, truncate(r.Str()))
 				}
 			}
 			modelpb.Labels(event.Labels).SetSlice(key, result)
--- a/input/otlp/metadata_test.go
+++ b/input/otlp/metadata_test.go
@@ -18,6 +18,7 @@
 package otlp_test
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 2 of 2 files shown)


marclop (approved):
Thanks!

1pkg (approved):

ericywl:
Is there anything that needs to be done aside from just pressing the merge button?

simitt:
>Is there anything that needs to be done aside from just pressing the merge button?

@ericywl we usually use `Squash and merge` (avoid creating a merge commit). Please ensure a good commit message, ideally including some term to close the related github card. If there is a reason for keeping several commits in the history, then you can use `Rebase and merge`. 
Other than that - just hit the button :)

carsonip:
> Is there anything that needs to be done aside from just pressing the merge button?

As apm-data is a library, to incorporate this change into the projects (e.g. apm-server), we'll need to cut a release (in this case, a patch release) and bump the apm-data depenedency version (which is usually done by automation) in the projects. I'm not sure if we have docs on this process. We can sync up if you want to give it a go.

ericywl:
> As apm-data is a library, to incorporate this change into the projects (e.g. apm-server), we'll need to cut a release (in this case, a patch release) and bump the apm-data depenedency version (which is usually done by automation) in the projects. I'm not sure if we have docs on this process. We can sync up if you want to give it a go.

Hey Carson, sure I will give it a try.

rubvs:
Tested manually:

Add the following after [Code](https://github.com/elastic/apm-server/blob/2e8498cd76be9606feda5e5b12d46a4568e766c2/systemtest/cmd/sendotlp/main.go#L189):
```go
parent.SetAttributes(attribute.StringSlice("keyvalue", []string{strings.Repeat("a", 20000)}))
```

Copy value from created `labels.keyvalue` in Kibana
Screenshot 2025-06-18 at 1 54 37 PM

Check that char length is `1024` in Bash:
```bash
> echo $#string_from_kibana
1024
```

PR #469, opened 2025-04-03
feat: ignore NaN float values

kruskall:
several clients are sending NaN in float values
ignore it to be more resilient even though it's not valid json

### DIFF
--- a/input/elasticapm/internal/modeldecoder/nullable/nullable.go
+++ b/input/elasticapm/internal/modeldecoder/nullable/nullable.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 	"unsafe"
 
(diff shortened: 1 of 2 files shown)


rubvs (approved):

carsonip (commented):
does "ignore NaN" mean silently dropping data without returning an error? Isn't the current behavior to fail explicitly better?

axw:
I agree with @carsonip. The marshaler can't make a decision about how to deal with this, so I think it should be handled in the client, e.g. return an error, log an error and continue, or silently drop the data.

axw:
@kruskall sorry I was a bit confusing above: when I wrote that comment I thought I was commenting on https://github.com/elastic/go-fastjson/pull/34 :sweat_smile:

Still, I agree that failing fast on NaN is preferable. I suggest we fix the thing that's producing metrics with NaN values. IIANM that's https://github.com/elastic/apm-agent-go/tree/main/module/apmprometheus

Issue #508, opened 2025-07-04
replace pkg/errors with stdlib


kruskall:
Similar to https://github.com/elastic/apm-server/issues/17478

https://github.com/pkg/errors/ was archived and is no longer maintained

stdlib has all the necessary methos (error wrapping, multierrors, etc.)

move to the stdlib and drop the dependency

PR #515, opened 2025-07-16
refactor: replace pkg/errors with stdlib

kruskall:
update generator to use stdlib instead of pkg/errors

regenerate models

Closes https://github.com/elastic/apm-data/issues/508


marclop (approved):
Nice. Another one bites the dust!

img

